PR #11, opened 2019-09-19
Add auto detection of region based on the license key

jorikvdwerf:


internal/cmd/scraper/scraper.go
```diff
@@ -59,7 +59,6 @@ func validateConfig(cfg *Config) error {
 
 // LoadViperDefaults loads the default configuration into the given Viper loader.
 func LoadViperDefaults(viper *viper.Viper) {
```
douglascamata: WDYT about moving this to the same file as the `loadConfig` function?

jorikvdwerf: good one! done

PR #15, opened 2019-09-25
Add support for summaries and histograms

alejandrodnm:
This PR adds support for processing summaries and histograms.

Most of the work of this PR was made by @MrAlias when the project was hosted on a private repository, so all the credit should go to him, I just rebased, and did some minor improvements.

The implementation follows the the [New Relic's guidelines for higher-level metric abstractions](https://github.com/newrelic/newrelic-exporter-specs/blob/master/Guidelines.md):

- For summaries, the quantiles are transformed and sent as percentiles.
- For histograms, predefined percentiles are calculated and sent along with the buckets. By default the 50, 95 and 99 percentile are calculated, this value can be overriden via the `percentiles` configuration option. 

The `histogram` package had to be copied from the official prometheus one, because the data structures and methods used are private in the official repo. Also, some changes were made to calculate percentiles instead of quantiles.

### DIFF
--- a/internal/histogram/percentile.go
+++ b/internal/histogram/percentile.go
@@ -116,7 +116,6 @@ func coalesceBuckets(buckets Buckets) Buckets {
 // As a somewhat hacky solution until ingestion is atomic per scrape, we
 // calculate the "envelope" of the histogram buckets, essentially removing
 // any decreases in the count between successive buckets.
-
 func ensureMonotonic(buckets Buckets) {
 	max := buckets[0].Count
 	for i := range buckets[1:] {
--- a/internal/histogram/percentile.go
+++ b/internal/histogram/percentile.go
@@ -1,0 +1,130 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package histogram
+
+import (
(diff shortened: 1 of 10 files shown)


internal/integration/emitter.go
```diff
@@ -6,18 +6,20 @@ package integration
...
 	"fmt"
-	"net/http"
-	"os"
+	"math"
 	"time"
 
 	"github.com/newrelic/go-telemetry-sdk/cumulative"
 	"github.com/newrelic/go-telemetry-sdk/telemetry"
+	"github.com/newrelic/nri-prometheus/internal/histogram"
+	mpb "github.com/prometheus/client_model/go"
```
douglascamata: Why **mpb**? I was reviewing the code and had no idea about what it meant. In other parts of the codebase we call this import `dto`.

internal/integration/emitter_test.go
```diff
@@ -83,7 +112,298 @@ func decodePromMetrics(src io.Reader) (*prometheus.MetricFamiliesByName, error)
 	return &mfs, nil
 }
 
-func TestNewTelemetryEmitterWithRoundTripper(t *testing.T) {
-	e := NewTelemetryEmitter("http://myapi.com", "myAPIkey", time.Second*20)
-	assert.Equal(t, "myAPIkey", e.apiKey)
```
douglascamata: Why removing this test?

internal/histogram/percentile.go
```diff
@@ -0,0 +1,130 @@
...
+// recent scrape.
+//
+// Monotonicity is usually guaranteed because if a bucket with upper bound
+// u1 has count c1, then any bucket with a higher upper bound u > u1 must
+// have counted all c1 observations and perhaps more, so that c  >= c1.
+//
+// As a somewhat hacky solution until ingestion is atomic per scrape, we
+// calculate the "envelope" of the histogram buckets, essentially removing
+// any decreases in the count between successive buckets.
+
```
douglascamata: Extra line break here?

internal/integration/emitter_test.go
alejandrodnm: because the key is not longer set in the `telemetryEmitter`.

PR #17, opened 2019-09-26
Documentation improvements and matching with the codebase

douglascamata:
* Better representation of the scrape timeout default value
* Fix the scrape timeout default value in the example configuration
* Fix the default value of `require_scrape_enabled_label_for_nodes` to match what is said in the default configuration

### DIFF
--- a/deploy/nri-prometheus.tmpl.yaml
+++ b/deploy/nri-prometheus.tmpl.yaml
@@ -92,7 +92,7 @@ data:
     insecure_skip_verify: false
     # The label used to identify scrapable targets. Defaults to "prometheus.io/scrape".
     scrape_enabled_label: "prometheus.io/scrape"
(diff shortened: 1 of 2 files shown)


deploy/nri-prometheus.tmpl.yaml
```diff
@@ -92,7 +92,7 @@ data:
     insecure_skip_verify: false
     # The label used to identify scrapable targets. Defaults to "prometheus.io/scrape".
     scrape_enabled_label: "prometheus.io/scrape"
-    # Whether k8s nodes needs to be labelled to be scraped or not. Defaults to false.
+    # Whether k8s nodes needs to be labelled to be scraped or not. Defaults to true.
```
alejandrodnm: Shouldn't it be `need` instead of `needs`

PR #19, opened 2019-10-08
Support for using a proxy for the telemetry emitter requests

alejandrodnm:
Adds proxy support for the telemetry emitter.

Adds three new config options `emitter_proxy`, `emitter_ca_file` and `emitter_insecure_skip_verify`.

The standard `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` couldn't be used for this purpose, because if set, they would also set the proxy for the k8s client that communicate with k8s api server.


deploy/nri-prometheus.tmpl.yaml
```diff
@@ -101,6 +101,34 @@ data:
     #      ca_file_path: "/etc/etcd/etcd-client-ca.crt"
     #      cert_file_path: "/etc/etcd/etcd-client.crt"
     #      key_file_path: "/etc/etcd/etcd-client.key"
+
+    # Proxy to use by the emitters when submitting metrics. It should be in the
+    # format [scheme]://[domain]:[port].
+    # The emitter is the component in charge of sending the scraped metrics.
+    # This proxy won't be used when scraping metrics from the targets.
+    # By default is empty, meaning that no proxy will be used.
```
jorikvdwerf:
```suggestion
    # By default it's empty, meaning that no proxy will be used.
```

deploy/nri-prometheus.tmpl.yaml
```diff
@@ -101,6 +101,34 @@ data:
     #      ca_file_path: "/etc/etcd/etcd-client-ca.crt"
     #      cert_file_path: "/etc/etcd/etcd-client.crt"
     #      key_file_path: "/etc/etcd/etcd-client.key"
+
+    # Proxy to use by the emitters when submitting metrics. It should be in the
```
jorikvdwerf:
```suggestion
    # Proxy to be used by the emitters when submitting metrics. It should be in the
```

deploy/nri-prometheus.tmpl.yaml
```diff
@@ -101,6 +101,34 @@ data:
...
+    # This proxy won't be used when scraping metrics from the targets.
+    # By default is empty, meaning that no proxy will be used.
+    # emitter_proxy: "http://localhost:8888"
+
+    # Certificate to add to the root CA that the emitter will use when
+    # verifying server certificates.
+    # If left empty, TLS uses the host's root CA set.
+    # emitter_ca_file: "/path/to/cert/server.pem"
+
+    # Wether the emitter should skip TLS verification when submitting data.
```
jorikvdwerf:
```suggestion
    # Whether the emitter should skip TLS verification when submitting data.
```

internal/cmd/scraper/scraper.go
```diff
@@ -40,6 +40,9 @@ type Config struct {
 	ProcessingRules                   []integration.ProcessingRule `mapstructure:"transformations"`
 	Percentiles                       []float64                    `mapstructure:"percentiles"`
 	DecorateFile                      bool
+	EmitterProxy                      string `mapstructure:"emitter_proxy"`
+	EmitterCaFile                     string `mapstructure:"emitter_ca_file"`
```
jorikvdwerf: could be useful to validate these config variables in the `validateConfig` function, e.g.: check if Proxy is set, and if so, if it's a valid URL. This way we can exit early.

internal/cmd/scraper/scraper.go
```diff
@@ -163,16 +166,44 @@ func Run(cfg *Config) {
...
+
+			transportOption, err := integration.TelemetryHarvesterWithInfraTransport(
+				cfg.LicenseKey,
+				cfg.EmitterProxy,
+				cfg.EmitterCaFile,
+				cfg.EmitterInsecureSkipVerify,
+			)
+
+			if err != nil {
+				logrus.Fatalf(
```
jorikvdwerf: I'm not a fan of exiting the application outside of the main() function. Could we return an error from the `run` function and quit outside of it?

internal/cmd/scraper/scraper.go
```diff
@@ -163,16 +166,44 @@ func Run(cfg *Config) {
...
 			harvesterOpts := []func(*telemetry.Config){
 				telemetry.ConfigAPIKey(cfg.LicenseKey),
 				telemetry.ConfigBasicErrorLogger(os.Stdout),
-				integration.TelemetryHarvesterWithInfraTransport(cfg.LicenseKey),
+				transportOption,
 				integration.TelemetryHarvesterWithMetricsURL(cfg.MetricAPIURL),
 				integration.TelemetryHarvesterWithHarvestPeriod(hTime),
 			}
 
+			logrus.Debugf("telemetry emitter configured with API endpoint: %s", cfg.MetricAPIURL)
```
jorikvdwerf: Outputting the same message twice. Also, this will spam the logs with a lot of `telemetry emitter configured with` messages; coming them into 1 message would be cleaner, like:
`logrus.Debugf("Telemetry emitter options: %#v", cfg)`

internal/integration/emitter.go
```diff
@@ -74,10 +75,34 @@ func TelemetryHarvesterWithHarvestPeriod(t time.Duration) TelemetryHarvesterOpt
 
 // TelemetryHarvesterWithInfraTransport wraps the `telemetry.Harvester`
 // `Transport` so that it uses the `licenseKey` instead of the `apiKey`.
-func TelemetryHarvesterWithInfraTransport(licenseKey string) TelemetryHarvesterOpt {
-	return func(cfg *telemetry.Config) {
-		cfg.Client.Transport = newInfraTransport(cfg.Client.Transport, licenseKey)
+func TelemetryHarvesterWithInfraTransport(
```
jorikvdwerf: Is there a way we can have 2 separate functions here, for example:
`TelemetryHarvesterWithLicenseKey` and `TelemetryHarvesterWithProxy`, and chain both together to have one http.RoundTripper? I'm not a fan of mixing both the licenseKey & proxy support into one thing.

alejandrodnm: done, I made the split between licenseKey, proxy and tlsConfig

internal/cmd/scraper/scraper.go
alejandrodnm: There is already a `logrus.Debugf("Telemetry emitter options: %#v", cfg)` line so I removed all the debugs.

cmd/nri-prometheus/main.go
```diff
@@ -14,5 +14,8 @@ func main() {
 		logrus.WithError(err).Fatal("while loading configuration")
 	}
 
-	scraper.Run(cfg)
+	err = scraper.Run(cfg)
+	if err != nil {
+		logrus.WithError(err).Fatal("execution finished with error")
```
jorikvdwerf: small thing, but finished indicates that it ran till completion, while it could not have been working at all

```suggestion
		logrus.WithError(err).Fatal("error occurred while running scraper")
```

or something like this

internal/integration/emitter.go
```diff
@@ -72,11 +75,70 @@ func TelemetryHarvesterWithHarvestPeriod(t time.Duration) TelemetryHarvesterOpt
...
+		rt := cfg.Client.Transport
+		if rt == nil {
+			rt = http.DefaultTransport
+		}
+
+		t, ok := rt.(*http.Transport)
+		if !ok {
+			logrus.Warning(
+				"telemetry emitter couldn't be configured with proxy, ",
+				"client transport is not and http.Transport, ",
```
jorikvdwerf:
```suggestion
				"client transport is not an http.Transport, ",
```

PR #21, opened 2019-10-18
Hide license key in logs

jorikvdwerf:


internal/integration/fetcher.go
```diff
@@ -279,7 +279,7 @@ const (
 	metricType_GAUGE     metricType = "gauge"
 	metricType_SUMMARY   metricType = "summary"
 	metricType_HISTOGRAM metricType = "histogram"
-	metricType_UNTYPED   metricType = "untyped"
+	//metricType_UNTYPED   metricType = "untyped"
```
lorgan3: is this intentional?

jorikvdwerf: Yep, it's an unused variable, the code won't compile with it on my machine. I'll just remove it instead of commenting it out

PR #22, opened 2019-10-21
FSI-1155: Don't scrape pods without a valid PodIP

jorikvdwerf:
This PR fixes the bug that adds newly created Pod-targets without an ipAddress and caches them, which makes them unscrabable unless you restart POMI. 

It also includes a simple program that runs the KubernetesTargetRetriever locally so you can debug it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,3 +52,16 @@ deploy/local.yaml.example to deploy/local.yaml and edit the placeholders.
    `https://newrelic.com/accounts/<YOUR_ACCOUNT_ID>`. It's located in the right sidebar.
 - After updating the yaml file, you need to compile the integration: `GOOS=linux make compile-only`.
 - Once you have it compiled, you need to deploy it in your Kubernetes cluster: `skaffold run`
+
+### Running the Kubernetes Target Retriever locally
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -52,3 +52,16 @@ deploy/local.yaml.example to deploy/local.yaml and edit the placeholders.
...
+
+It can be useful to run the Kubernetes Target Retriever locally against a remote/local cluster to debug the endpoints that are discovered.
+The program located in `/cmd/k8s-target-retriever` is made for this.
+
+To run the program, go to `cmd/k8s-target-retriever` in your terminal, and run the following command:
+```shell script
+# ensure your kubectl is configured correcly & against the correct clusters
+kubectl config get-contexts
+# run the program 
+go run * -kubeconfig ~/.kube/config
```
douglascamata: (forgot to press the button here)

Could we run this tool from the root of the project? Having to cd in and out of this folder to run it can be annoying.

jorikvdwerf: sure, done!

jorikvdwerf:
@alejandrodnm done!

PR #31, opened 2019-11-28
Fix: NaN values produce invalid json

alejandrodnm:
Some metrics which are NaN cause the sdk to produce an invalid json, the marshaling of the json doesn't fail and it's a problem in the telemetry sdk. I already open an issue in their repo.

https://github.com/newrelic/go-telemetry-sdk/issues/2

This invalid json was being sent to newrelic, and because it couldn't be parsed, the whole metrics batch was discarded. 

This PR checks the values for NaN or Inf before sending them to the sdk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Fixed
+- Check gauges and counter values for NaN or Inf, before emitting them with the
+  telemetry sdk. There is an issue with the telemetry sdk were sending NaN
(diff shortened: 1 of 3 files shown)


jorikvdwerf:
Hm, if the library has been fixed, we will not need to have this check in place, right? Can we try to have them fix it asap (or even do it for them), and not add this additional check?

alejandrodnm:
@douglascamata actually this NaNs won't be sent. Do you think we should send them as nulls? what do you think @jorikvdwerf.

I really don't know the implications or the UX of sending them as null.

I already ping the sdk team for an ETA to see if they are going to fix it on their end.

douglascamata:
I would expect them to be present as null because the metrics are technically there, but only have this NaN value for the moment. To me it sounds different from not having the metrics at all.

alejandrodnm:
This was fixed directly in the SDK 

https://github.com/newrelic/newrelic-telemetry-sdk-go/pull/3

PR #36, opened 2020-01-31
Upgrade telemetry sdk

lorgan3:
Remove the old vendored telemetry-sdk and replace it with the current open sourced one.
I also had to do some small changes to the code because some error handling has changed.

Also removed the current NaN code because the telemetry sdk handles it now and fix an issue with headers.


internal/cmd/scraper/scraper.go
```diff
@@ -244,7 +244,12 @@ func Run(cfg *Config) error {
 				Percentiles:   cfg.Percentiles,
 				HarvesterOpts: harvesterOpts,
 			}
-			emitters = append(emitters, integration.NewTelemetryEmitter(c))
+
+			emitter, err := integration.NewTelemetryEmitter(c)
+			if err != nil {
+				return err
```
jorikvdwerf: For debugging it's helpful to annotate the errors, to give a bit of context, e.g.: `return errors.Wrap(err, "could not create  new TelemetryEmitter`)`

internal/integration/roundtripper.go
```diff
@@ -11,7 +11,7 @@ type licenseKeyRoundTripper struct {
 // RoundTrip wraps the `RoundTrip` method removing the "X-Insert-Key"
```
jorikvdwerf: outdated comments

PR #46, opened 2020-05-26
Fix bug with URL encoded params

douglascamata:
The manual construction of the struct `url.URL` was considering the URL
parameters as a part of the URLs's path and this was causing encoding
issues on some characters.

As an example, if a pod had the `prometheus.io/path` annotation set to
`/metrics?format=prometheus` we would end up querying
`/metrics%3Fformat=prometheus`.

### DIFF
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -178,17 +178,14 @@ func objectTargets(object metav1.Object) []Target {
 func serviceTarget(s *apiv1.Service, port, path string) Target {
 	lbls := labels.Set{}
 	hostname := fmt.Sprintf("%s.%s.svc", s.Name, s.Namespace)
-	addr := url.URL{
-		Scheme: "http",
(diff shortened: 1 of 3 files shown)


internal/pkg/endpoints/kubernetes.go
```diff
@@ -178,17 +178,14 @@ func objectTargets(object metav1.Object) []Target {
 func serviceTarget(s *apiv1.Service, port, path string) Target {
 	lbls := labels.Set{}
 	hostname := fmt.Sprintf("%s.%s.svc", s.Name, s.Namespace)
-	addr := url.URL{
-		Scheme: "http",
-		Host:   net.JoinHostPort(hostname, port),
-		Path:   path,
-	}
+	hostAndPort := net.JoinHostPort(hostname, port)
+	addr, _ := url.Parse(fmt.Sprintf("http://%s%s", hostAndPort, path))
```
jorikvdwerf, line 182: why are you ignoring this error?

PR #49, opened 2020-06-05
Update project to use go modules

areina:
In the past wasn't possible to use go modules because it didn't work well with k8s client-go library. However, there is already good support for it in newer versions of that library and this project seems to work well with modules.


alejandrodnm:
Snyk is failing in the CI

alejandrodnm:
We are setting the `SNYK_TOKEN` and the service account is still valid, can you trigger the CI again?

douglascamata:
@alejandrodnm I retriggered the build and and it failed the same way. 🤔

alejandrodnm:
I told @areina that I will look at this, it seems to be a problem I'll try with a different service account.

areina:
Here you have the reason why the build is failing: https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

We should find another way of integrating the project with snyk. Did we consider using the Snyk's GH integration instead of calling snyk test from travis?

alejandrodnm:
@areina The snyk docs say that they only support go modules via the snyk CLI https://support.snyk.io/hc/en-us/articles/360003817417-Snyk-for-Golang

alejandrodnm:
@areina I'll ask snyk support but I think we will have to disable snyk for PRs where the env var is not set, at least this way when the PR is merged the merged job will run and snyk will run.

areina:
@alejandrodnm while you get an answer from Snyk, I've created a new PR #51 skipping `snyk test` on PR builds. If they provide another solution, we could get back the current approach.

PR #60, opened 2020-07-28
Vendor dependencies, move tests/linting to Github actions

jorik:
... and update/add open source files


.github/workflows/security.yaml
```diff
@@ -0,0 +1,15 @@
+name: Security Scan
+
+on: [push, pull_request]
```
douglascamata, line 3: We lint and test also on tags to version releases. Can we also do security scans there too?

PR #63, opened 2020-08-13
feat(integrationBehaviour): the integration accept a new parameter to…

paologallinaharbur:
The nri-prometheus integration now accepts new config/parameters:
 - **StandAlone** in the config file permits changing the whole behaviour of the integration
 - **ConfigPath** argument allows to set a new path for the configuration file used by Viper (previously hardcoded). In particular that feature is going to be used by the agent when passing the path to a tmp file generated from the integration config file.

>By default the behaviour is the old one, however if standAlone is set to True a new behaviour kicks in.

Moreover the Viper version has been bumped up since the file passed by the agent is extensionless and only a newer version of Viper supports it and `go mod tidy` has been run

**Old behaviour:**
 - nri-prometheus is capable to run as a standAlone integration, it manages automatically scraping metrics, processing them and sending data to the NewRelic. 
The process does not terminate after sending the data, therefore once defined the targets they get scraped periodically and the result is sent
Moreover the integration is exposing telemetry data through a prometheus endpoint. 

**New behaviour**
 - Currently it is just a stub (printing "New Behaviour" and exiting) in order to allow reviewing the change of the main in an easier way.
   The new behaviour would transform this integration making it more similar to a onHost integration, therefore it will focus merely on scraping data and processing it. The execution and sending the data will be managed by the agent.

   It will output the data on stdOut and exit leveraging the agent to send data and to perform auto-discovery


alejandrodnm:
Hi @paologallinaharbur can you please explain the old and new behaviours in the PR and why it's needed

cmd/nri-prometheus/config.go
```diff
@@ -4,25 +4,46 @@ package main
...
 
+	"github.com/newrelic/infra-integrations-sdk/args"
 	"github.com/newrelic/nri-prometheus/internal/cmd/scraper"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 )
 
+// ArgumentList Available Arguments
+type ArgumentList struct {
+	ConfigPath string `default:"test" help:"Path to the config file"`
```
davidbrota, line 21: Can't we use viper to read the flag?

configs/nri-prometheus-config.yml.sample
```diff
@@ -0,0 +1,20 @@
+integrations:
+  - name: com.newrelic.prometheus
+    exec:
+      - /var/db/newrelic-infra/newrelic-integrations/bin/nri-prometheus
+    config:
```
davidbrota, line 5: wouldn't be better to use this?
```
- name: nri-prometheus
  config:
      ...
```
we don't need to specify the whole path, the agent will search for the executable in some known folders and we use the usual naming convention for the inventory 
https://docs.newrelic.com/docs/create-integrations/infrastructure-integrations-sdk/specifications/host-integrations-newer-configuration-format#inventory_source

configs/nri-prometheus-config.yml.sample
```diff
@@ -0,0 +1,20 @@
...
+      # The name of your cluster. It's important to match other New Relic products to relate the data.
+      cluster_name: "mq_exporter"
+
+      # Points to the exporter launched in docker-compose
+      targets:
+        - description: "mq exporter"
+          urls: ["http://127.0.0.1:8115"]
+      verbose: true
+      insecure_skip_verify: true
+      metric_api_url: https://staging-metric-api.newrelic.com/metric/v1/infra
```
davidbrota, line 15: we should remove the staging url

cmd/nri-prometheus/config.go
```diff
@@ -54,6 +75,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("emitter_harvest_period", "1s")
 	viper.SetDefault("auto_decorate", false)
 	viper.SetDefault("insecure_skip_verify", false)
+	viper.SetDefault("stand_alone", true)
```
davidbrota, line 78: standalone should be false by default (we want POMI to behave as an on-host integration by default)
also, I would put it together "standalone", but ask Fabrizio about this.

configs/nri-prometheus-config.yml.sample
```diff
@@ -0,0 +1,20 @@
...
+
+      stand_alone: false
```
davidbrota, line 17: this is not needed if standalone is false by default. In this case, we should update the current K8S deployments to set standalone as true

cmd/nri-prometheus/config.go
paologallinaharbur: I'll update the default!
@theletterf What do you think about the config name? I considered it being a two words expression

configs/nri-prometheus-config.yml.sample
paologallinaharbur: I'll update k8s config manifests

cmd/nri-prometheus/config.go
davidbrota: also, the standalone flag should also work as environment variable

paologallinaharbur: However I am bit worried on making as the default a new behaviour.

If they are running the integration outside K8s with a config file and they update the integration it will break since the standalone config will be not there.

theletterf: Let's go with "standalone" https://brians.wsu.edu/2016/05/31/standalone-stand-alone

paologallinaharbur: I'll add it to the dockerfile,  ENV VARS were binding already with Viper

cmd/nri-prometheus/config.go
paologallinaharbur: Yes but in accepting args we can build a similar experience to other integration, es:
 - `help` command with documentation
 - we should add common commands, es `version`
Standard errors:
```
flag provided but not defined: -config_path2
Usage of nri-prometheus:
  -config_path string
        Path to the config file (default "test")
```

configs/nri-prometheus-config.yml.sample
```diff
@@ -0,0 +1,27 @@
...
+      #    tls_config:
+      #      ca_file_path: "/etc/etcd/etcd-client-ca.crt"
+      #      cert_file_path: "/etc/etcd/etcd-client.crt"
+      #      key_file_path: "/etc/etcd/etcd-client.key"
+
+      # Wether the integration should run in verbose mode or not. Defaults to false.
+      verbose: false
+
+      # The HTTP client timeout when fetching data from endpoints. Defaults to 30s.
+      # scrape_timeout: "30s"
```
douglascamata, line 20: Maybe it's a good idea to remove this. It doesn't make sense if the integration is being executed by the agent and die as soon as it finishes its work.

cmd/nri-prometheus/config.go
```diff
@@ -4,25 +4,46 @@ package main
...
 
+// ArgumentList Available Arguments
+type ArgumentList struct {
+	ConfigPath string `default:"" help:"Path to the config file"`
+}
+
 func loadConfig() (*scraper.Config, error) {
+
+	c := ArgumentList{}
+	err := args.SetupArgs(&c)
```
douglascamata, line 27: Please test that using the infra integrations sdk args it's not interfering with Viper.

configs/nri-prometheus-config.yml.sample
paologallinaharbur: Why? I removed the  `scrape_duration: "30s"` since the integration dies, but the timeout isn't still useful in case the scraper does not respond?

douglascamata: Ah, I confused this with the interval, sorry @paologallinaharbur. 😓

cmd/nri-prometheus/config.go
paologallinaharbur: I checked the code and ArgumentList does not interfere with Viper.
I also tested manually locally and in the container I got from the Dockerfile included

configs/nri-prometheus-config.yml.sample
```diff
@@ -0,0 +1,27 @@
+integrations:
+  - name: nri-prometheus
+    inventory_source: integration/com.newrelic.prometheus
```
gsanchezgavier, line 3: I think we could remove this and use the default instead. `inventory_source` is set by default as integration/nri-prometheus . [ docs ](https://docs.newrelic.com/docs/create-integrations/infrastructure-integrations-sdk/specifications/host-integrations-newer-configuration-format#inventory_source).

jorik: Do you know why it was set explicitly here, @gsanchezgavier? Removing it here and using the default will change the actual name.

gsanchezgavier: no i don't, but inventory is not being used afaik here.

morecar (approved):
koolio

configs/nri-prometheus-config.yml.sample
jorik: I'm just keeping it like this for now, and discuss this with Paolo when he's back. We'll not do a release with it yet, anyway.

PR #64, opened 2020-08-14
The integration should scrape metrics, process them and exit

paologallinaharbur:
**Disclaimer**

This main aim of this PR is add a new behaviour in case `nri-prometheus` is executed as onHost integration.
In this case the integration is **not** long-running and should scrape data, process them and exit

**New features:**
 - when running in `standalone=false` mode the integration runs the scrape process only once, process the data and exit
 - Added some tests to cover Happy path of the scraper and exporter not answering
 - If the exporter answer with a status code not in the 200 range an error is logged and the body is not parsed

**Decisions:**
 - does it make sense to keep `scrape_duration` logic (the integration distributes the scraping of the targets in that timeframe) in this use-case? 
I decided to keep it in order to reduce the size of the changes lowering in the config file the default scrape_duration to 5s

**Todo Comments:**
 - The current code is using the stdOut emitter (that do not print anything since the fields of the structure are not exported), a following PR should add the logic to add each metric to an integration sdkv4 object.
 - Once updated the test can double-check the scraped data as well


cmd/nri-prometheus/main.go
```diff
@@ -20,6 +20,11 @@ func main() {
 			logrus.WithError(err).Fatal("error occurred while running scraper")
 		}
 	} else {
-		logrus.Info("New behaviour!")
+		// todo create a proper emitter that add metrics to an integration and prints them to stdout once scraping and processing is performed
+		cfg.Emitters = []string{"integrationSDK4"}
```
ardias, line 24: i don't think this is a good way to do it. I think we should still use the configuration file to add the emitters.
I guess the idea is to force to use only one (the infra sdk one) but I don't think it's really required. The standalone function and the configuration file will be managed by the Agent so I would leave this to the configuration

paologallinaharbur: Sure, this is just a stub waiting for an other piece of work that will connect scraper and emitter in the proper way.
Moreover emitter type is configurable but only one option makes sense when standalone=false.

ardias (approved):

morecar (approved):
LGTM

configs/nri-prometheus-config.yml.sample
```diff
@@ -19,6 +18,9 @@ integrations:
       # The HTTP client timeout when fetching data from endpoints. Defaults to 30s.
       # scrape_timeout: "30s"
 
+      # Length in time to distribute the scraping from the endpoints.
+      scrape_duration: "5s"
```
gsanchezgavier, line 22: I think this two configs will have duplicated functionality with `interval` and `timeout` from the Agent and in the future we should find a way to just keep the Agent ones

gsanchezgavier (approved):
lgtm

configs/nri-prometheus-config.yml.sample
ardias: yes. What I'm not 100% sure yet is if we are going to keep the "standalone" behaviour and so we would need these options, of if we are completely replacing the way POMI works to only work with the agent

ardias:
This is changing the default behaviour so it means it's a breaking change. We have to release with a major version bump.

ardias:
Made a small change to make standalone=true the default to keep backwards compatibility

morecar (approved):

configs/nri-prometheus-config.yml.sample
```diff
@@ -1,10 +1,9 @@
 integrations:
   - name: nri-prometheus
-    inventory_source: integration/com.newrelic.prometheus
     config:
       # The name of your cluster. It's important to match other New Relic products to relate the data.
       cluster_name: "my_exporter"
-      
+
```
gsanchezgavier, line 6: `standalone: false` needs to be added in this config

gsanchezgavier (approved):
looks good thanks!, just need to add the standalone=false to config for agent

ardias (approved):

PR #67, opened 2020-08-24
feat(emitter): add infra sdk emitter

ardias:
Adds the Infra sdk emitter to the list of possible emitters. The Infra SDK emitter will output to stdout the metrics in the new SDK V4 format which is used for dimensional metrics support.


ardias:
@carlosroman Added you as reviewer to keep you in the loop of what's being changed in POMI. Ofc suggestions welcome

ardias:
We have a dependency with https://github.com/newrelic/nri-prometheus/pull/64 in order to work best. It might need a rebase/merge after that PR gets merged

internal/cmd/scraper/scraper.go
```diff
@@ -311,11 +290,21 @@ func Run(cfg *Config) error {
 				return errors.Wrap(err, "could not create new TelemetryEmitter")
 			}
 			emitters = append(emitters, emitter)
+		case "infra-sdk":
```
gsanchezgavier, line 293: do we need to keep "stdout" emitter? i think that @paologallinaharbur mentioned that was not working since the struct fields were not exporter.

gsanchezgavier (approved):
lgtm! thanks

internal/cmd/scraper/scraper.go
paologallinaharbur: When using it the output is `{"","","","",""}`, however it is used to run test. 
I would keep it, even if the same is quite confusing (mock emitter would be better I guess)

internal/integration/emitter.go
```diff
@@ -4,19 +4,9 @@
 package integration
```
paologallinaharbur, line 4: We have the scraper in a separate package, does it make sense to separate emitters as well?
Does it require a lot of work?

internal/integration/infra_sdk_emitter.go
```diff
@@ -0,0 +1,181 @@
...
+func (e *InfraSdkEmitter) emitSummary(i *integration.Integration, metric Metric, timestamp time.Time) (err error) {
+	summary, ok := metric.value.(*dto.Summary)
+	if !ok {
+		return fmt.Errorf("unknown summary metric type for %q: %T", metric.name, metric.value)
+	}
+
+	m, err := integration.Summary(timestamp, metric.name+"_sum",
+		1,
+		math.NaN(),             // can we calc average here?
+		summary.GetSampleSum(), // not sure the delta gets calculated in the agent for summary metrics
```
paologallinaharbur, line 143: should we verify this comment before merging?

internal/integration/infra_sdk_emitter_test.go
```diff
@@ -0,0 +1,260 @@
...
+# HELP redis_exporter_scrapes_total Current total redis scrapes.
+# TYPE redis_exporter_scrapes_total counter
+redis_exporter_scrapes_total{hostname="localhost",env="dev"} 3
+`
+	// when they are scraped
+	metrics := scrapeString(t, input)
+	return metrics.Metrics
+}
+
+//---- simplified structs mimicking the real Infra SDK output structure
```
paologallinaharbur, line 229: Nice approach!

paologallinaharbur (dismissed):

internal/integration/infra_sdk_emitter.go
```diff
@@ -0,0 +1,181 @@
...
+// InfraSdkEmitter is the emitter using the infra sdk to output metrics to stdout
+type InfraSdkEmitter struct {
+	deltaCalculator *cumulative.DeltaCalculator
+}
+
+// NewInfraSdkEmitter creates a new Infra SDK emitter
+func NewInfraSdkEmitter() (*InfraSdkEmitter, error) {
+	return &InfraSdkEmitter{
+		// perhaps allow configuration..
+		deltaCalculator: cumulative.NewDeltaCalculator(),
```
paologallinaharbur, line 25: This works when the integration is long-running, does it work as well if it terminates?
Does it stores the values on file?

internal/integration/emitter.go
ardias: shouldn't be too hard. I'll change it

internal/integration/infra_sdk_emitter.go
ardias: I think we can remove them. It was copied basically from the other emitter so regarding the average I think we could, but the spec for the NR prometheus "consumer" did not include it: https://source.datanerd.us/core-data-platform/prometheus-consumer#conversion-rules
Regarding the second comment, I'm still not sure if it will work if we do nothing. I don't think the agent will do anything, it will only calculate deltas for counters. But the telemetry emitter also does not calculate the delta so let's keep it as is and when we get everything working check if the values are correct

internal/integration/infra_sdk_emitter.go
ardias: No it does not store anything in files, it's memory only. 
It's probably not going to work very well because we're going to run this only once so the values in memory will be lost.
I don't have a good approach to fix this. Only thing I can think of is using the SDK "storer" to keep the value in a file and replicate the implementation of the deltaCalculator. 
What do you think?

paologallinaharbur: We could also read the stored values and "simulate" we receive those metrics in order to initialise the deltaCalculator. Otherwise I have the feeling that that the deltas will not work as expected

gsanchezgavier: couldn't we cherry pick the storer implementatino from the master sdk branch and use that? I guess that the sdk storer is going to be needed for the integrations already using it.

ardias: I believe we removed it because the agent would be responsible to keep that data to calculate deltas. I'll recover the storer implementation, that's probably easier to do.

ardias: I've added an implementation here. But it's possible that this will be done in the agent.

gsanchezgavier (approved):

internal/integration/infra_sdk_emitter.go
paologallinaharbur: I do not see any new commit, are you going to include the implementation in this PR?

ardias: @paologallinaharbur sorry, i had it in another branch and didn't merge it to my repo main. You should see it now

ardias: Removed that commit because we've agreed that the delta calculation will be done in the agent. We'll still need a few changes because the way we send the histogram and summary metrics is most likely going to change

paologallinaharbur (approved):

paologallinaharbur (approved):
Thanks for the change!

paologallinaharbur:
There are conflicting files and no test can run

ardias:
> There are conflicting files and no test can run, can you fix that? A part from that since last review the only change I see are the updated dependencies and the vendor folder

The last changes were just updating to the correct infra-sdk version fixing the conflict with main. It's fixed

paologallinaharbur (approved):

PR #68, opened 2020-08-26
feat: add spec files for entity modeling

gsanchezgavier:
This PR adds the spec files model that will be use to compose entities with the metrics based on the file definition. 

The functions added here will be used by the integration when running with the Agent and using sdk emitter.

All spec files will be located in a specific folder and pomi will load them every execution.


internal/integration/spec.go
```diff
@@ -0,0 +1,114 @@
...
+	"path"
+	"regexp"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+const fileNameMatcher = `^prometheus_.*\.yml$`
```
ardias, line 18: small comment here: some yaml files sometimes are created with `.yaml` extension so we could also include them automatically instead of forcing renaming to `.yml`

internal/integration/spec.go
```diff
@@ -0,0 +1,114 @@
...
+			entityType = metricType
+			for _, d := range e.Properties.Dimensions {
+				var val interface{}
+				var ok bool
+				// the metric needs all the dimensions defined to avoid entity name collision
+				if val, ok = m.attributes[d]; !ok {
+					return "", "", fmt.Errorf("dimension %s not found in metric %s", d, m.name)
+				}
+				// entity name will be composed by the value of the dimensions defined for the entity in order
+				entityName = entityName + ":" + fmt.Sprintf("%v", val)
```
ardias, line 106: not sure this is going to work ok for most cases. Not 100% sure also if we want to allow for configurable entity names by the user.  This would mean a user might want a certain format for their entity names and another user a different format. In the end it's us that either approve or change the naming format so perhaps it's a good idea to allow for a dynamic format for the name. wdyt?

internal/integration/spec.go
```diff
@@ -0,0 +1,114 @@
...
+// Spec contains the rules to group metrics into entities
+type Spec struct {
+	Provider string      `yaml:"provider"`
+	Service  string      `yaml:"service"`
+	Entities []EntityDef `yaml:"entities"`
+}
+
+// EntityDef has info related to each entity
+type EntityDef struct {
+	Type       string        `yaml:"name"`
```
ardias, line 34: why not Name instead of Type?

ardias: I guess Type works because it really is the entity type and not its name. The field in the spec file is not very well "named"

internal/integration/spec.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		logrus.Debugf("spec file loaded for service:%s", sd.Service)
+		specs.SpecsByName[sd.Service] = sd
+	}
+
+	return specs, nil
+}
+
+// getEntity returns entity name and type of the metric based on the spec configuration defined for the service.
+// metric example: serviceName_entityName_metricName{dimension="dim"} 0
```
ardias, line 79: This won't work very well in some cases. For example, in redis we have a metric named **redis_last_slow_execution_duration_seconds** which won't translate very well. The "entity name" here would be "last_slow_execution" but this code will name it "last".
I think we probably need to declare a format or something that can work with more metrics

internal/integration/spec.go
gsanchezgavier: yes I followed the sdk naming here where it is named as type

internal/integration/spec.go
gsanchezgavier: yes this will not work for all the cases, i talked with @morecar about this and we decided to go for this case and put the rest of the metrics in a main entity for each service, so right now all metrics that do not match with an entity described like this will go to the main entity.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/stretchr/objx v0.1.2-0.20180626195558-9e1dfc121bca // indirect
 	github.com/stretchr/testify v1.6.1
 	golang.org/x/crypto v0.0.0-20200220183623-bac4c82f6975 // indirect
+	gopkg.in/yaml.v2 v2.2.8
```
paologallinaharbur, line 26: have you run as well `go mod vendor`? Because I do not see the new dependency vendored

internal/integration/spec.go
```diff
@@ -0,0 +1,114 @@
...
```
paologallinaharbur, line 79: How do we assure that there are no collision between different exporters? Es Redis and Mysql having both the entity `database`. shouldn't we create the entity type as "serviceName.entityName" ?

gsanchezgavier: good catch, I could use the same approach cloud integrations use provider+serviceName+entityName

go.mod
gsanchezgavier: yes , it is already there and in go.sum . it's wired that it was not listed as indirect here.

internal/integration/spec.go
```diff
@@ -0,0 +1,147 @@
...
+	return specs, nil
+}
+
+// getEntity returns entity name and type of the metric based on the spec configuration defined for the service.
+// conditions for the metric:
+//   - metric.name has to start a prefix that matches with a service from the spec files
+//   - metric.name has to be defined in one of the entities of the spec file
+//   - if dimension has been specified for the entity, the metric need to have all of them.
+//   - metrics that belongs to entities with no dimension specified will share the same name
+func (s *Specs) getEntity(m Metric) (entityName string, entityType string, err error) {
```
gsanchezgavier, line 90: this is a helper function that will be used by the infra sdk emitter. metrics that return err will be added to the LocalEntity for now.

internal/integration/spec.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Specs contains all the services specs mapped with the service name
+type Specs struct {
+	SpecsByName map[string]SpecDef
+}
+
+// SpecDef contains the rules to group metrics into entities
+type SpecDef struct {
+	Provider string      `yaml:"provider"`
+	Service  string      `yaml:"service"`
```
paologallinaharbur, line 28: I had the some doubt when writing the prometheus parser, I do not know if we can assume we have only one service. 
As we discussed today we might have a single exporter returning two (or more "services"). In that case either we will have multiple file or we need to change the structure to something like:

```
type SpecDef struct {
    Provider string      `yaml:"provider"`
    Service  []Service `yaml:"service"`
}

type Service struct {
	ServiceName  string           `yaml:"service"`
	Entities            []EntityDef  `yaml:"entities"`
}
```
In this case we would not follow cloud integration structure

paologallinaharbur (approved):
LGTM!

internal/integration/spec.go
gsanchezgavier: Thanks for bring this up. If for some reason the same exporter exposes metrics from different services we can define a spec file for each of the service and keep the same structure. 
The service will be picked from the metric prefix and treated as defined in the corresponding spec file.
I'm thinking about that this solution could also work for the exporter related metrics like go_ or process_ :)

ardias: @gsanchezgavier To be clear, you're saying that in the case of 2 or more services for the same exporter we will create a file for each one? If that is your idea, I just worry about how it will impact external contributors, especially customers. It may be difficult to explain that, even with documentation, and the more files we have the harder it gets to contribute. We should really try to reduce the number of files required to make an exporter work.
I'm not sure we need to follow the exact same structure of the cloud integrations, especially if it doesn't work well for us.
Just my 2cts

internal/integration/spec.go
ardias: Maybe for later, but if you expect that there will be different errors being returned, maybe we should have "typed" errors (like ErrServiceNotFound, ErrEntityNotFound, etc...)

internal/integration/spec.go
paologallinaharbur: I agree that we should try as much as possible to have _one_ file per integration. 

Either we follow the cloud integration approach adding more services with the `---`
ES:
```
provider: prometheus
service: ravendb1
display_name: Raven Db
entities:
[...]
---
provider: prometheus
service: ravendb2
display_name: Raven Db
entities:
[...]
```
Or we can change the structure allowing arrays of services in one yaml. We will need to change a bit the automatic tool but then we will be free to diverge when needed

ES:
```
provider: prometheus
display_name: Raven Db
services:
  - ravendb1
     entities:
     [...]
  - ravendb2
     entities:
     [...]
```

internal/integration/spec.go
gsanchezgavier: I was thinking not to use those errors more than for printing for debug proposes.

internal/integration/spec.go
gsanchezgavier: @morecar  needs your opinion here. I think cloud int. also splits some specs files for some integrations. 
I also want to mention that i think having 2 different services in one exporter don't looks like a good prometheus [best practice](https://prometheus.io/docs/practices/naming/#metric-names).

ardias: I think we may need to look into some more exporters to get an idea of the number of "services" that they expose.
On the other hand we can just assume that 1 exporter = 1 service. For example Redis has metrics that start with very different "prefixes": **redis**_ , **process**_ , **go**_ . If we ignore the **go** prefix (which the exporter can be configured to not return), the metrics all belong to the Redis service, just with different entities.  I'm not sure about other exporters but I'm guessing it may be the same.

In summary, I would not make services an array. If we find we need more than one service for one exporter we can use the first option, having multiple services on the same file

ardias (approved):

internal/integration/spec.go
paologallinaharbur: sounds good to me

paologallinaharbur (approved):
LGTM.

internal/integration/spec_test.go
```diff
@@ -0,0 +1,152 @@
...
+		{
+			name:   "matchEntityWithoutDimensions",
+			fields: fields{specs.SpecsByName},
+			args: args{
+				Metric{
+					name:       "ravendb_document_put_bytes_total",
+					attributes: labels.Set{},
+				}},
+			wantEntityName: "node",
+			wantEntityType: "PrometheusRavendbNode",
```
paologallinaharbur, line 67: Are we going to remove `Prometheus` prefix from entity name in next iteration?

gsanchezgavier: yes i'll do it

gsanchezgavier:
rebased from main and remove provider prefix

paologallinaharbur (approved):
Thanks!

PR #69, opened 2020-08-28
feat(config): added config option to enable/disable autodiscovery

paologallinaharbur:
**Use Case**
Sometimes we want to run the integration in an environment where autodiscovery is not needed or due to permission limitation cannot be performed. In this scenarios hundreds of error logs were shown together with failing request to k8s APIs.

Es: service account running the pod ha not visibility into the whole cluster.
```
time="2020-08-28T11:20:53Z" level=warning msg="couldn't list node resource, retrying" component=KubernetesAPI error="timeout reached, 60 retries executed. last error: nodes is forbidden: User \"system:serviceaccount:new-relic:nri-prometheus\" cannot list resource \"nodes\" in API group \"\" at the cluster scope"
```

**Solution**
 - A config option `enable_autodiscovery ` has been added in order to add the possibility to enable/disable autodiscovery.
 - By default such config is set to `true`

UPDATE

After discussion the config option as been called `disable_autodiscovery` with default `false`

### DIFF
--- a/cmd/nri-prometheus/config.go
+++ b/cmd/nri-prometheus/config.go
@@ -76,7 +76,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("auto_decorate", false)
 	viper.SetDefault("insecure_skip_verify", false)
 	viper.SetDefault("standalone", true)
-	viper.SetDefault("enable_autodiscovery", true)
+	viper.SetDefault("disable_autodiscovery", false)
 	viper.SetDefault("percentiles", []float64{50.0, 95.0, 99.0})
 }
 
--- a/cmd/nri-prometheus/config.go
+++ b/cmd/nri-prometheus/config.go
@@ -76,6 +76,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("auto_decorate", false)
 	viper.SetDefault("insecure_skip_verify", false)
 	viper.SetDefault("standalone", true)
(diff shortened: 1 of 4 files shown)


cmd/nri-prometheus/config.go
```diff
@@ -76,6 +76,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("auto_decorate", false)
 	viper.SetDefault("insecure_skip_verify", false)
 	viper.SetDefault("standalone", true)
+	viper.SetDefault("enable_autodiscovery", true)
```
ardias, line 79: shouldn't it be "**disable**_autodiscovery"?

ardias (changes requested):

cmd/nri-prometheus/config.go
paologallinaharbur: it is the same, but `disable_autodiscovery` default would be `false` and we would check if we need to add targets with: 
```
if !disable_autodiscovery{
}
```

isn't that less clear? do we usually avoid to use `enable_feature` having true as default?

paologallinaharbur: I am changing it, I thought about it and I guess you are right it is more clear for the user with the disable

paologallinaharbur: @ardias Updated :)

ardias (approved):

PR #70, opened 2020-08-28
fix(k8sManifest): configfile added as alias of config_path

paologallinaharbur:
**Reason behind the change**

In the k8s manifests I noticed that the `--configfile ` argument was passed. However, such argument was ignored since the config file path was hardcoded.

After we moved to the infraSDK if an unexpected argument is passed an error is returned. Therefore after updating the manifest with the correct argument I added the old argument as an alias.

Be aware that we decided to change the config name since the agent pass the environment variable `CONFIG_PATH`

### DIFF
--- a/cmd/nri-prometheus/config.go
+++ b/cmd/nri-prometheus/config.go
@@ -19,6 +19,7 @@ import (
 // ArgumentList Available Arguments
 type ArgumentList struct {
 	ConfigPath string `default:"" help:"Path to the config file"`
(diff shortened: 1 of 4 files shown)


paologallinaharbur:
A different possibility would be to accept only the `configfile` argument and link manually the Environment Variable CONFIG_PATH passed by the agent. I do not like especially any of the solutions

ardias (approved):

PR #73, opened 2020-09-01
Ignored Nan and Infinite float values

jorik:
These values cannot be handled by NRDB, and POMI currently logs an error message for every single NaN/Infinite value it processes, which can be confusing.

### DIFF
--- a/internal/integration/emitter.go
+++ b/internal/integration/emitter.go
@@ -4,6 +4,7 @@
 package integration
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
@@ -30,10 +31,16 @@ type Emitter interface {
 	Emit([]Metric) error
 }
 
+// Harvester aggregates and reports metrics and spans
+type harvester interface {
+	RecordMetric(m telemetry.Metric)
+	HarvestNow(ct context.Context)
+}
+
 // TelemetryEmitter emits metrics using the go-telemetry-sdk.
 type TelemetryEmitter struct {
 	name            string
-	harvester       *telemetry.Harvester
+	harvester       harvester
 	deltaCalculator *cumulative.DeltaCalculator
 }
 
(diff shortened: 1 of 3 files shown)


internal/integration/harvester_decorator.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func (ha harvesterDecorator) HarvestNow(ctx context.Context) {
+	ha.innerHarvester.HarvestNow(ctx)
+}
+
+func (ha harvesterDecorator) processMetric(f float64, m telemetry.Metric) {
+	if isNaNOrInfinity(f) {
+		logrus.Debugf("Ignoring NaN float value for metric: %v", m)
```
paologallinaharbur, line 36: it could be also ±inf

internal/integration/emitter.go
```diff
@@ -161,14 +168,19 @@ func NewTelemetryEmitter(cfg TelemetryEmitterConfig) (*TelemetryEmitter, error)
...
-	harvester, err := telemetry.NewHarvester(cfg.HarvesterOpts...)
+	var h harvester
+	h, err := telemetry.NewHarvester(cfg.HarvesterOpts...)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not create new Harvester")
 	}
 
+	// Wrap the harvester so we can filter out invalid float values: NaN and Infinity.
+	// If we do send them, the harvester will always output these as errors
+	h = harvesterDecorator{h}
```
paologallinaharbur, line 179: nice approach!

paologallinaharbur (approved):
LGTM! thanks for the change!

internal/integration/harvester_decorator.go
jorik: fixed!

paologallinaharbur (approved):
LGTM, Thanks for the update!

PR #74, opened 2020-09-03
chore: cleanup some cruft and move some stuff around

ardias:


internal/integration/emitter.go
```diff
@@ -12,6 +12,8 @@ import (
 const (
 	defaultDeltaExpirationAge           = 5 * time.Minute
 	defaultDeltaExpirationCheckInterval = 5 * time.Minute
+	instrumentationName                 = "nri-prometheus"
+	instrumentationVersion              = "2.0.1"
```
paologallinaharbur, line 16: Should we link this variable to a config at build time?

`-v -ldflags '-X main.buildVersion=$(VERSION)'`

Otherwise it will be often outdated

paologallinaharbur (dismissed):
LGTM
I left only a minor comment regarding making the version dynamic or not; we could either tackle that in this PR or open a specific one for that logic.

internal/integration/emitter.go
ardias: Oops, i think we probably don't need that. I just saw right now where we are setting the version to the integration, so we don't need these 2 consts

paologallinaharbur (approved):

PR #77, opened 2020-09-04
feat(emitter): infra sdk emitter now maps metrics to entities

ardias:
Maps metrics to the proper entities. Uses the integration definition files for the mapping

### DIFF
--- a/deploy/local.yaml.example
+++ b/deploy/local.yaml.example
@@ -117,6 +117,8 @@ data:
     #          container: "containerName"
     #          pod: "podName"
     #          deployment: "deploymentName"
+    # integration definition files required to map metrics to entities
+    definition_files_path: /etc/newrelic-infra/definition-files
 kind: ConfigMap
 metadata:
   name: nri-prometheus-cfg
--- a/deploy/local.yaml.example
+++ b/deploy/local.yaml.example
@@ -118,7 +118,7 @@ data:
     #          pod: "podName"
     #          deployment: "deploymentName"
     # integration definition files required to map metrics to entities
-    definition_files_path: /etc/newrelic-infra/definition-files
+    # definition_files_path: /etc/newrelic-infra/definition-files
 kind: ConfigMap
 metadata:
   name: nri-prometheus-cfg
--- a/internal/cmd/scraper/scraper.go
+++ b/internal/cmd/scraper/scraper.go
@@ -49,6 +49,7 @@ type Config struct {
 	EmitterInsecureSkipVerify                    bool          `mapstructure:"emitter_insecure_skip_verify" default:"false"`
 	TelemetryEmitterDeltaExpirationAge           time.Duration `mapstructure:"telemetry_emitter_delta_expiration_age"`
 	TelemetryEmitterDeltaExpirationCheckInterval time.Duration `mapstructure:"telemetry_emitter_delta_expiration_check_interval"`
+	DefinitionFilesPath                          string        `mapstructure:"definitions_file_path"`
 }
 
 const maskedLicenseKey = "****"
@@ -293,7 +294,11 @@ func Run(cfg *Config) error {
 			}
 			emitters = append(emitters, emitter)
 		case "infra-sdk":
-			emitter := integration.NewInfraSdkEmitter()
+			specs, err := integration.LoadSpecFiles(cfg.DefinitionFilesPath)
+			if err != nil {
+				return err
+			}
+			emitter := integration.NewInfraSdkEmitter(specs)
 			emitters = append(emitters, emitter)
 		default:
 			logrus.Debugf("unknown emitter: %s", e)
--- a/internal/cmd/scraper/scraper.go
+++ b/internal/cmd/scraper/scraper.go
@@ -49,7 +49,7 @@ type Config struct {
 	EmitterInsecureSkipVerify                    bool          `mapstructure:"emitter_insecure_skip_verify" default:"false"`
 	TelemetryEmitterDeltaExpirationAge           time.Duration `mapstructure:"telemetry_emitter_delta_expiration_age"`
 	TelemetryEmitterDeltaExpirationCheckInterval time.Duration `mapstructure:"telemetry_emitter_delta_expiration_check_interval"`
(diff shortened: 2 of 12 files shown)


gsanchezgavier (approved):
looks good thanks!

deploy/local.yaml.example
```diff
@@ -117,6 +117,8 @@ data:
     #          container: "containerName"
     #          pod: "podName"
     #          deployment: "deploymentName"
+    # integration definition files required to map metrics to entities
```
paologallinaharbur, line 120: Shouldn't in k8s this be commented? The default behaviour should be `standalone` so `DefinitionFilesPath` is not used, no?

paologallinaharbur (approved):
Everything looks good, I only have a doubt regarding the k8s config

internal/integration/infra_sdk_emitter.go
```diff
@@ -122,8 +118,30 @@ func (e *InfraSdkEmitter) emitSummary(i *integration.Integration, metric Metric,
...
+	if err != nil {
+		i.HostEntity.AddMetric(m)
+		return nil
+	}
+
+	// try to find the entity and add the metric to it
+	// if there's no entity with the same name yet, create it and add it to the integration
+	entity, ok := i.FindEntity(entityName)
+	if !ok {
+		entity, err = i.NewEntity(entityName, entityType, entityName)
```
gsanchezgavier, line 136: we will need the host:port of the target somewhere here , not sure if we could add it after

deploy/local.yaml.example
ardias: if we are packaging POMI with the agent it will always be standalone: false and we will need a path to the definition files, no? Even in K8s this is how it's going to work I think, right?

paologallinaharbur: Yes, once it is packaged with the agent that is going to be the configuration, but currently since it is not if `standalone: false` the pod created by the deployment will start and exit each time entering in a `crashloopbackoff` state

ardias:
I've added the host of the target to the entity name to make them distinct, although I think it would be better to have some kind of "format" capability to generate the entity name in the definition file.
For example we could have something like:
`
entities:
  - name instance:${dimension:hostname}
  - name: database:${cluster}:${dimension:db}
`
This would make the entity name generation to be more flexible and not have to be defined in the code. And in the code the same rule applies to all entities which may not be what we want. Just a though for a later feature perhaps

paologallinaharbur (dismissed):
LGTM

deploy/local.yaml.example
ardias: ok I see your point. I'll change it back for now

internal/integration/infra_sdk_emitter.go
```diff
@@ -122,11 +120,66 @@ func (e *InfraSdkEmitter) emitSummary(i *integration.Integration, metric Metric,
 		ps.AddQuantile(*q.Quantile, *q.Value)
 	}
 
-	i.HostEntity.AddMetric(ps)
+	return e.addMetricToEntity(i, metric, ps)
+}
+
+func (e *InfraSdkEmitter) addMetricToEntity(i *integration.Integration, metric Metric, m metrics.Metric) error {
```
jorik, line 126: what is the difference between `metric Metric` and `m metrics.Metric`? They could use some more descriptive names

internal/integration/spec.go
```diff
@@ -47,6 +48,14 @@ type MetricDef struct {
 	Name string `yaml:"provider_name"`
 }
 
+type entityNameProps struct {
```
paologallinaharbur, line 51: can you add just some info regarding this struct?
I understand why is there but in the future it might be confusing

paologallinaharbur (dismissed):
LGTM!
I checked only diff from the last review, since I reviewed the rest before

internal/integration/infra_sdk_emitter.go
```diff
@@ -122,11 +120,66 @@ func (e *InfraSdkEmitter) emitSummary(i *integration.Integration, metric Metric,
...
+		i.AddEntity(entity)
+	}
 
+	entity.AddMetric(m)
 	return nil
 }
 
+// build the entity name based on various properties
+// the format should be as follows:
+//  serviceName:exporterHost:exporterPort:entityName:dimension1:dimension2..
```
paologallinaharbur, line 154: is there a limit for the length of the entity name we can hit?

internal/integration/infra_sdk_emitter.go
ardias: I know. there is a *metric* type from POMI and there's a *metric* type from the Infra SDK. I'll add some prefixes to the imports to make it more clear

internal/integration/infra_sdk_emitter.go
ardias: good point. I think there is. we should confirm this with platform

ardias: Platform has told me that it's 500 chars. So we should perhap introduce a limit here around that value.
I see an issue where if we remove some dimensions from the name it can lead to the entity name not being unique.

paologallinaharbur (approved):
Thanks for the changes!

paologallinaharbur (approved):

gsanchezgavier (dismissed):

internal/cmd/scraper/scraper.go
```diff
@@ -49,6 +49,7 @@ type Config struct {
 	EmitterInsecureSkipVerify                    bool          `mapstructure:"emitter_insecure_skip_verify" default:"false"`
 	TelemetryEmitterDeltaExpirationAge           time.Duration `mapstructure:"telemetry_emitter_delta_expiration_age"`
 	TelemetryEmitterDeltaExpirationCheckInterval time.Duration `mapstructure:"telemetry_emitter_delta_expiration_check_interval"`
+	DefinitionFilesPath                          string        `mapstructure:"definitions_file_path"`
```
gsanchezgavier, line 52: typo `mapstructure:"definitions_file_path"` "definition_files_path"

paologallinaharbur (approved):

gsanchezgavier (approved):

Issue #83, opened 2020-09-23
Issues when not running in standalone mode


davidbrota:
when `standalone` is `false`:

- the default emitter should be `infra-sdk`
- the `definition_files_path` config option should have a default value 
- if you don't provide the definition files, a debug log message shows up for every metric

```
time="2020-09-23T14:22:12Z" level=debug msg="time=\"2020-09-23T14:22:12Z\" level=debug msg=\"failed to map metric to entity. using 'host' entity\" error=\"no spec files for service: ravendb\"" component=integrations.runner.Group integration_name=nri-prometheus
```

which spams the debug log and it's quite annoying. We should only log once per namespace.

gsanchezgavier:
the PR has been merged

PR #85, opened 2020-10-06
fix: default configs emitter and definition path

gsanchezgavier:
Adds default value for:
* DefinitionFilesPath
* Emitter: uses `infra-sdk` when standalone is false and `telemetry` when is true (backward compatibility)

This PR fixes issue #83

### DIFF
--- a/internal/cmd/scraper/scraper.go
+++ b/internal/cmd/scraper/scraper.go
@@ -296,7 +296,7 @@ func Run(cfg *Config) error {
 		case "infra-sdk":
 			specs, err := integration.LoadSpecFiles(cfg.DefinitionFilesPath)
 			if err != nil {
-				return err
+				logrus.Errorf("error loading definition files: %s", err)
 			}
 			emitter := integration.NewInfraSdkEmitter(specs)
 			emitters = append(emitters, emitter)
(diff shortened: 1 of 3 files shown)


paologallinaharbur (dismissed):
LGTM!

internal/cmd/scraper/scraper.go
```diff
@@ -296,7 +296,7 @@ func Run(cfg *Config) error {
 		case "infra-sdk":
 			specs, err := integration.LoadSpecFiles(cfg.DefinitionFilesPath)
 			if err != nil {
-				return err
+				logrus.Errorf("error loading definition files: %s", err)
```
paologallinaharbur, line 299: Do we want to carry on even if the file is not found/not loaded?

gsanchezgavier: i don't see why not. When the scraped target has no matching on the definitions the metrics are added to the host entity. Also this is checking that the folder is there

paologallinaharbur: Yes that is right 👍

paologallinaharbur (approved):
thanks for the changes

ardias (approved):

PR #86, opened 2020-10-06
Various performance improvements & adjustable fetcher worker threads

jorik:


internal/cmd/scraper/scraper.go
```diff
@@ -97,6 +95,11 @@ func validateConfig(cfg *Config) error {
 		}
 	}
 
+	if cfg.WorkerThreads < 4 {
```
paologallinaharbur, line 98: Does it make sense to set as well a maximum? Es 1000?

cmd/nri-prometheus/config.go
```diff
@@ -83,6 +83,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("standalone", true)
 	viper.SetDefault("disable_autodiscovery", false)
 	viper.SetDefault("percentiles", []float64{50.0, 95.0, 99.0})
+	viper.SetDefault("worker_threads", 4)
```
gsanchezgavier, line 86: why not to increase the default value , since this was causing issues in large environments?

internal/integration/fetcher.go
```diff
@@ -176,7 +181,8 @@ func (pf *prometheusFetcher) Fetch(targets []endpoints.Target) <-chan TargetMetr
 				Info("Target list for fetching metrics is empty")
 			return
 		}
-		ticker := time.NewTicker(pf.duration / time.Duration(nTargets))
+
+		ticker := time.NewTicker((pf.duration / 2) / time.Duration(nTargets))
```
paologallinaharbur, line 185: why half of the duration?

internal/cmd/scraper/scraper.go
jorik: I don't think it's be necessary, but we should add a comment explaining what the setting does.
The min-check was mostly to prevent people from accidentally setting it to zero

internal/integration/fetcher.go
jorik: I just pushed a new commit explaining why!

paologallinaharbur (dismissed):
Everything looks good, only small comments

paologallinaharbur:
I do not know how difficult it is, but it would nice to expose in the `/metrics` per each worker how much time it took to scrape the targets and how much time to process and send data.

But it might be out of scope

cmd/nri-prometheus/config.go
ardias: it's usually enough. the problem is when the target takes more time to respond it blocks everything.

ardias:
@paologallinaharbur it already does. We have metrics for those, they start with nr_stats_*

ardias (approved):

paologallinaharbur (approved):
LGTM

PR #90, opened 2020-10-10
Improve CA support configuration for HTTPS scrapping

bartoszj:
When configuring `CA_FILE` (like in case of Kubernetes setup) it is not possible to scrape TLS/HTTPS endpoints because there is only once valid certificate. This change is using standard default CA certificates and adds provided CA certificate to the pool. Standard certificates can still be configured by using `SSL_CERT_FILE` and `SSL_CERT_DIR`.


paologallinaharbur:
Stale PR, closing.

wu-json:
@paologallinaharbur  Can we revive this one? Having some trouble getting `nri-prometheus` to work with an external target url with mTLS using self-signed certificates. I've been getting "x509 signed by unknown authority" errors and this should fix those: https://forfuncsake.github.io/post/2017/08/trust-extra-ca-cert-in-go-app/

PR #94, opened 2020-10-16
feat: accept duplicated service definitions

gsanchezgavier:
This PR allows two exporter with same metric namespace to have different definition files . 

This approach just adds the entities from all definitions files that share the same service name to the same object. 

In the future the entity discovery mechanism should be replaced and this approach will no longer be needed.

### DIFF
--- a/internal/integration/spec.go
+++ b/internal/integration/spec.go
@@ -91,7 +91,14 @@ func LoadSpecFiles(filesPath string) (Specs, error) {
 			continue
 		}
 		logrus.Debugf("spec file loaded for service:%s", sd.Service)
-		specs.SpecsByName[sd.Service] = sd
+		if spec, ok := specs.SpecsByName[sd.Service]; ok {
+			spec.Entities = append(spec.Entities, sd.Entities...)
+			specs.SpecsByName[sd.Service] = spec
+			//TODO how to manage defaultEntity for each spec?
+		} else {
+			specs.SpecsByName[sd.Service] = sd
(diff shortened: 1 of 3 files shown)


paologallinaharbur (approved):

internal/integration/spec.go
```diff
@@ -91,7 +91,14 @@ func LoadSpecFiles(filesPath string) (Specs, error) {
 			continue
 		}
 		logrus.Debugf("spec file loaded for service:%s", sd.Service)
-		specs.SpecsByName[sd.Service] = sd
+		if spec, ok := specs.SpecsByName[sd.Service]; ok {
+			spec.Entities = append(spec.Entities, sd.Entities...)
+			specs.SpecsByName[sd.Service] = spec
```
ardias, line 96: just curiosity, do you have to re-assign the variable in the map? Wouldn't it be sufficient to just append the entities?

ardias (approved):

internal/integration/spec.go
gsanchezgavier: It was complaining that it was possible , but I will dig into that

gsanchezgavier: It's because the map references to a value and not a pointer. And there's no guarantee on the location of that value. I'll change that to a pointer

https://stackoverflow.com/questions/32751537/why-do-i-get-a-cannot-assign-error-when-setting-value-to-a-struct-as-a-value-i

PR #101, opened 2020-10-23
fix(summary): summary "sum" metric must also be sent as delta

### LINKED ISSUE #100: Summary metric should do delta calculation for summary metric

## Description
Due to the "sum" not being delta calculated like "count" we cannot perform a proper average over boths values.

## Expected Behavior

"sum" value in the summary metric should be delta calculated. See histogram for how it should be done

ardias:
in order to properly allow for average calculation, the "sum" attribute of the summary must also be delta

fixes #100


internal/integration/telemetry_sdk_emitter.go
```diff
@@ -238,15 +238,17 @@ func (te *TelemetryEmitter) emitSummary(metric Metric, timestamp time.Time) erro
...
-		Min:        math.NaN(),
-		Max:        math.NaN(),
-		Timestamp:  timestamp,
-	})
+	if sumCount, ok := te.deltaCalculator.CountMetric(metric.name+"_sum", metric.attributes, float64(summary.GetSampleSum()), timestamp); ok {
+		te.harvester.RecordMetric(telemetry.Summary{
+			Name:       metric.name + "_sum",
+			Attributes: metric.attributes,
+			Count:      1,
+			Sum:        sumCount.Value,
```
gsanchezgavier, line 246: I would thought that this value already come in the original prometheus metric and we just need it to send it as it is.

gsanchezgavier (approved):

internal/integration/telemetry_sdk_emitter.go
ardias: In prometheus it comes as a counter and so in order for us to be able to calculate the average (_sum/_count) we also need to send it as a delta, like we do for the _count

PR #109, opened 2020-11-30
mockexporter: go implementation of a mock exporter

### LINKED ISSUE #105: Load test exporter mock
[Modify](https://github.com/newrelic/nri-prometheus/blob/main/deploy/static/static.yaml)/Create the exporter Mock  to set the timeout configurable and handle multiple requests concurrently (each of these exporters will be mapped to multiple services and will receive concurrent connection from POMI)  and different set of metrics from a file.

roobre:
Mockexporter is a simple http server that echoes metrics loaded from a file on disk. It features several parameters useful for laod testing such as custom simulated latency, and limitation of concurrency.

### Example `yml`

An example YML is provided for "quick&dirty" usage. It uses `golang:alpine` and downloads the mock exporter script from this repo, then `go run`s it. It also downloads a metrics file if the appropriate environment variable is set.

### Docker image

A Dockerfile is available, and a prebuilt image can be found at [roobre/mockexporter](https://hub.docker.com/r/roobre/mockexporter). A sample file can be mounted and then referenced using the `-metrics` flags or the `METRICS` env var:

    docker run -v `realpath custom.metrics.file`:/app/different.metrics.file roobre/mockexporter -metrics different.metrics.file

### Arguments and configuration

mockexporter receives its arguments either from command line flags, or as environment variables.

Environment vars have the same name as the command line flags, but `UPPERCASED` and with dashes `-` replaced with underscores `_` so that, for example, the `latency-variation` flag could also be read from the `LATENCY_VARIATION` env var.

The full list of configuration options can be seen in the very beginning of the code.

---

Closes #105.


paologallinaharbur:
Can we add as well a "big" prometheus payload?

roobre:
> Can we add as well a "big" prometheus payload?

It would be ideal to do so, but I think we can use the `METRICS_FILE_URL` "feature" of the k8s yaml for that in order to not clog the repo with sample data.

deploy/static/static.yaml
```diff
@@ -21,18 +21,24 @@ spec:
...
       containers:
         - name: from-file-prometheus-exporter
-          image: python:alpine3.9
+          image: golang:alpine
           env:
             - name: METRICS_FILE_URL
-              value: "<ADD_YOUR_FILE_URL_HERE>" # You can use private gist url.
+              value: "" # You can use private gist url.
+            - name: TESTER_SCRIPT_URL
+              value: "https://raw.githubusercontent.com/newrelic/nri-prometheus/santalla/mockexporter/internal/mockexporter/loadtest.go".
```
roobre, line 29: Change this to `main` branch before merging

roobre:
I haven't tested the yaml file on k8s yet (I still haven't set up a local cluster...), please feel free to report all the funny ways it will crash.

gsanchezgavier:
there are some files that has prometheus metrics used by test already in the repo, also we could pick from a gh gist. I would not upload a metrics file for this at this point, but we probably need it when automating the test

deploy/static/static.yaml
```diff
@@ -21,18 +21,24 @@ spec:
...
             - name: metrics
               containerPort: 8080
           command: ["/bin/sh","-c"]
           # The reason of using a URL instead of a config map is that the latest has a limitation of up to 1MB
-          args: ["wget $METRICS_FILE_URL -O /etc/from-file-prometheus-exporter/metrics; python -m http.server -b 0.0.0.0 -d /etc/from-file-prometheus-exporter/ 8080"]
+          args:
+            - |
+              [[ -n "$METRICS_FILE_URL" ]] && wget $METRICS_FILE_URL -O /app/metrics
+              [[ -n "$TESTER_SCRIPT_URL" ]] && wget $TESTER_SCRIPT_URL -O /app/loadtest.go
+              go run /app/loadtest.go -addr :8080 -metrics /app/metrics -latency 50 -latency-variation 15
```
gsanchezgavier, line 39: not sure if this is going to work outside the gopath without go.mod file.

roobre: It does (at least locally) as a kind of an edge case, because all dependencies are from the standard library. It will probably be messier if it had to download some.

internal/mockexporter/metrics.file
```diff
@@ -0,0 +1,10 @@
+# HELP http_request_duration_seconds request duration histogram
```
gsanchezgavier, line 1: is this file needed? I guess we will need a larger number of metrics when implement the automated test

roobre: I just copied a random example for the net for this. My idea was to keep using the metrics download functionality to feed a huge file from a gist/pastebin without polluting the repo with it.

roobre: Update: I checked on a local k8s cluster and it seems to work well :)

deploy/static/static.yaml
```diff
@@ -21,25 +21,44 @@ spec:
...
         - name: from-file-prometheus-exporter
-          image: python:alpine3.9
-          env:
+          image: golang:alpine
+          env: # Files will be downloaded from the following URLs if they are not empty
             - name: METRICS_FILE_URL
-              value: "<ADD_YOUR_FILE_URL_HERE>" # You can use private gist url.
+              value: "https://raw.githubusercontent.com/newrelic/nri-prometheus/santalla/mockexporter/internal/mockexporter/metrics.file"
+            - name: TESTER_SCRIPT_URL
+              value: "https://raw.githubusercontent.com/newrelic/nri-prometheus/santalla/mockexporter/internal/mockexporter/loadtest.go"
```
paologallinaharbur, line 29: Now I get what you meant during the daily, it is the first time I see this approach 😄 
I prefer the Dockerfile to build and image and use that locally, but if it is works let's keep this :)

internal/mockexporter/metrics.file
gsanchezgavier: ok so this file could be removed right?

internal/mockexporter/loadtest.go
```diff
@@ -0,0 +1,92 @@
+package main
+
+import (
```
paologallinaharbur, line 3: I would not place this under internal. Usually we do not place there main packages.

When we have these we place them in `./cmd/DifferentCMD`. Should we move these files in a new folder there?

internal/mockexporter/loadtest.go
```diff
@@ -0,0 +1,92 @@
...
+	"flag"
```
paologallinaharbur, line 4: I would use github.com/namsral/flag instead of standard library, 
it allows using config trough environment variable that is quite useful with containers

roobre: Point taken, now config files can be read from env vars as well. In order to keep the `.go` file "dependency-less", I just made a simple wrapper for `flag` instead.

internal/mockexporter/loadtest.go
roobre: I agree that `internal` is not the best place for it, however I'm also not sure of whether this is useful enough externally to give it its own subdir in cmd. I don't really have a preference for neither option. What are @gsanchezgavier and @arvdias thoughts on this?

deploy/static/static.yaml
roobre: Dockerfile is up now as well!

internal/mockexporter/metrics.file
roobre: Removed in https://github.com/newrelic/nri-prometheus/pull/109/commits/f6fb74cd52f8282f660e44792ed4872c50ca00e1

gsanchezgavier (approved):

internal/mockexporter/loadtest.go
gsanchezgavier: I would put it under /tools and if we do the automated load test move it to /test imho

paologallinaharbur: One of the CMD I believe it is used already for debugging only, but yes tools or test work as well 😃

roobre: I moved this to `internal/cmd` since it already existed. I have zero problems moving it into `/tools` if you think it's a better fit:
🚀 : Move to `tools`
👍 : Keep in `internal/cmd`

gsanchezgavier (approved):

paologallinaharbur (approved):

paologallinaharbur (approved):

PR #111, opened 2020-12-02
chore: Add removed gha

gsanchezgavier:
Adds GithubActions removed in #103  and replaces the secrets to org level ones


cmd/nri-prometheus/versioninfo.json
```diff
@@ -1,44 +0,0 @@
-{
-	"FixedFileInfo":
-	{
```
paologallinaharbur, line 3: Why removing this? Asking just in case

gsanchezgavier: Not sure exactly what this file is, but it was accidentally included in the [commit](https://github.com/newrelic/nri-prometheus/pull/103) , and i checked that the pipeline remove it in the [clean step](https://github.com/newrelic/nri-prometheus/blob/9ace92f5754afe4dfc5c0e82d2385d5ac8476475/build/release.mk#L18)

.github/workflows/helm.yml
```diff
@@ -0,0 +1,19 @@
...
+  release:
+    types: [released]
+
+jobs:
+  create_helm_chart_pr:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Create PR using Version Bump 
+      run: |
+        appVersion=${{github.event.release.tag_name}}
```
roobre, line 12: I'm sorry I just could not hold myself
```suggestion
        appVersion=${{ github.event.release.tag_name }}
```

.github/workflows/helm.yml
```diff
@@ -0,0 +1,19 @@
...
+        chartName=nri-prometheus
+        
+        curl -H "Accept: application/vnd.github.everest-preview+json" \
+        -H "Authorization: token ${{ secrets.VERSION_BUMP_TOKEN }}" \
```
roobre, line 16: This is probably good as it is, but I wanted to double check: Would this work with `secrets.GITHUB_TOKEN`, which is automatically created by github actions? Maybe it won't if it's a different repo, but might be worth saving yet another secret to keep track of.

.github/workflows/prerelease.yml
```diff
@@ -122,8 +122,8 @@ jobs:
       - name: Login to DockerHub
         uses: docker/login-action@v1 
         with:
-          username: ${{ secrets.DOCKER_USERNAME}}
-          password: ${{ secrets.DOCKER_PASSWORD}}      
+          username: ${{ secrets.FSI_DOCKERHUB_USERNAME}}
+          password: ${{ secrets.FSI_DOCKERHUB_TOKEN}}
```
roobre, line 126:
```suggestion
          username: ${{ secrets.FSI_DOCKERHUB_USERNAME }}
          password: ${{ secrets.FSI_DOCKERHUB_TOKEN }}
```

.github/workflows/push_pr.yml
```diff
@@ -0,0 +1,96 @@
+name: Push/PR pipeline
+
+on:
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
```
roobre, line 8: For other pipelines we use
```suggestion
on:
  push:
    tags:
      - v*
    branches:
      - main
  pull_request:
```
Might be a good opportunity to change it

.github/workflows/helm.yml
gsanchezgavier: sure i just check out this file from the last commit it was available. :)

.github/workflows/helm.yml
gsanchezgavier: I worked on this one and sadly it needs a Personal token kind :( . [More details](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows)

roobre (approved):
I haven't checked strictly whether this correspond exactly to what it was deleted but otherwise LGTM.

PR #112, opened 2020-12-03
Load Test chart

### LINKED ISSUE #106: Load test environment Helm chart
Create a chart that will generate multiple services using a few exporters mocks pods:
The service created will be labeled as prometheus endpoints so POMI can automatically discover them and scrape them. Since these are much lighter than pods we can create hundreds of them in a local environment and simulate targets that can be auto-discovered.
The prometheus labels for auto discovery `prometheus.io/port`, `prometheus.io/scrape` should be added to the services fo auto-discovery functionality.

paologallinaharbur:
Fix #106 

This PR add a chart to generate prometheus load. 

It creates `n` services per deployment where `n` is `numberServicesPerDeploy`
The deployments can be configured with the array:
```
deployments:
  - name: one
    latency: "12"
    latencyVariation: "0"
    metrics: "https://raw.githubusercontent.com/newrelic/nri-prometheus/main/internal/cmd/scraper/testData/testData.prometheus"
    #maxRoutines: ""
```
Notice that each values has default and that none of them is required to be unique:

Name of deployment: `<name>-lat<latency>-latvar<latencyVar>-<deployindex>`

Name of service: `<name>-lat<latency>-latvar<latencyVar>-<deployindex>-<serviceindex>`

### DIFF
--- a/load-test/Chart.yaml
+++ b/load-test/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+name: load-test
+description: A Helm chart for prometheus load generator.
+type: application
+version: 0.1.0
+appVersion: 0.1.0
--- a/load-test/.helmignore
+++ b/load-test/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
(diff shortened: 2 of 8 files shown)


load-test/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: load-test
+description: A Helm chart for Kubernetes
```
gsanchezgavier, line 3: could you add a short description that explains the objective of simulate targets

gsanchezgavier (approved):
great work! thanks

load-test/Chart.yaml
paologallinaharbur: I'll add as well a short readme

gsanchezgavier (approved):

PR #117, opened 2020-12-12
Reworked example transformation section, added worker_threads, and isolated defaults

bpschmitt:
Changes:

- Moved default settings to their own line so that it's easier to find
- Added `worker_threads`
- Modified example `transformations` to exclude metrics collected by our Daemonset

### DIFF
--- a/deploy/nri-prometheus.tmpl.yaml
+++ b/deploy/nri-prometheus.tmpl.yaml
@@ -83,42 +83,53 @@ data:
     # The name of your cluster. It's important to match other New Relic products to relate the data.
     cluster_name: "<YOUR_CLUSTER_NAME>"
 
-    # When standalone is set to false nri-prometheus requires an infrastructure agent to work and send data. Defaults to true
+    # When standalone is set to false nri-prometheus requires an infrastructure agent to work and send data. 
+    # Default: true
     # standalone: true
 
-    # How often the integration should run. Defaults to 30s.
+    # How often the integration should run. 
+    # Default: "30s"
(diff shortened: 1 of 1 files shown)


deploy/nri-prometheus.tmpl.yaml
```diff
@@ -83,40 +83,55 @@ data:
...
+    # Default: "prometheus.io/scrape"
     scrape_enabled_label: "prometheus.io/scrape"
 
-    # Whether k8s nodes need to be labelled to be scraped or not. Defaults to true.
+    # Whether k8s nodes need to be labelled to be scraped or not. 
+    # Default: true
     require_scrape_enabled_label_for_nodes: true
 
+    # Number of worker threads used for scraping targets.
+    # For large clusters with many endpoints, increase to 10.
```
gsanchezgavier, line 131: @bpschmitt we are currently working on a load test and posible refactor to improve performance related to the number of workers. We have detected that increasing this number without reducing the `emitter_harvest_period` could lead to a high increase of memory consumption. Also the core consumption peak is something that needs to be taken into account. We plan to write a general recommendation for this configuration.

roobre:
Hi @bpschmitt, thank you very much for improving this file. As @gsanchezgavier commented, we made some changes to the config files related to the number of workers and max amount of metrics to store in memory.

Could you rebase this changes and check if there is any conflicting information? Thanks!

paologallinaharbur:
Stale, closing for now, feel free to reopen :)

Issue #108, opened 2020-11-30
Workers load test


gsanchezgavier:
* For fixed amount of targets and CPU increase the number of workers and chart the total time the scraping.
Run a test with a high number of targets and of workers. Understand which is safe range for the number of workers given different CPU configurations (assuming no latencies)
* Document suggestion for the number of workers recommended for amount of targets. Add target scrape duration check in troubleshoot guide to identify targets with big latency that could block workers.

gsanchezgavier:
Docs added 
https://docs.newrelic.com/docs/integrations/prometheus-integrations/troubleshooting/sparse-data-missing-metrics-data-gaps

PR #118, opened 2020-12-14
Implement boundedHarvester

### LINKED ISSUE #121: Increasing the number of workers causes memory usage to skyrocket

## Description
Increasing the number of workers to overcome network latency causes nri-prometheus' memory usage to skyrocket to several gigabytes.

## Expected Behavior
Memory usage should remain reasonable regarding the number of workers

## Additional context
This is caused by the Telemetry SDK harvester not harvesting quick enough the metrics scraped by the workers. A workaround could be to reduce the Harvesting period to a very small amount, however that would needlessly put load on the client and server even on scenarios with less workers.

This was discovered as a part of #108

roobre:
## Background

During the worker load testing activities from #108, we found that the increasing the number of workers was the most feasible solution for scaling `nri-prometheus` so it can handle thousands of targets in less than a minute. Additionally, increasing the number of workers makes the integration more tolerant to slow exporters, needing more of them to completely block the scraping process.

However, increasing the number of targets and workers brought a huge memory increase. A following investigation pointed the Harvester as the source of this increase: Metrics start piling up in memory waiting to be harvested, which can cause huge memory spikes.

Reducing the harvest period is a valid workaround for this, however it is not ideal, as it does not scale with the real load. This means that the hand-picked time can bee too much for small loads (tenths of targets) and at the same time not enough to big ones (more than several thousands), while also putting too much stress on the backend.

## Proposal

This PR implements `boundedHarvester`, a harvester implementation and wrapper that keeps count of the number of metrics that are waiting to be harvested. `boundedHarvester` will very often check if the number of accumulated metrics is above a given threshold, and trigger a harvest if that happens.

### Gotchas

This functionality could also be implemented in the Telemtry SDK itself, probably providing an overall simpler solution. However, since it might be out of scope for the sdk, the wrapper approach aims to be faster to deploy.

## Pending tasks

- [x] Basic implementation
- [x] Manual testing with the load testing infra
- [x] Unit tests

---

Fixes #121


internal/integration/bounded_harvester.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/newrelic/newrelic-telemetry-sdk-go/telemetry"
+	"github.com/sirupsen/logrus"
+)
+
+const boundHarvesterDefaultPeriod = 3 * time.Second
```
paologallinaharbur, line 12: the current Default is 1s

internal/integration/bounded_harvester.go
```diff
@@ -0,0 +1,114 @@
...
+	reportedMetrics int
+	lastReport      time.Time
+
+	inner harvester
+}
+
+// RecordMetric records the metric in the underlying harvester and reports all of them if needed
+func (h *boundedHarvester) RecordMetric(m telemetry.Metric) {
+	h.inner.RecordMetric(m)
+	h.reportIfNeeded(context.Background(), 1, false)
```
paologallinaharbur, line 74: does it make sense to check this for each metric?
each time we need to wait on the mux check and carry on, we could check each MinReportInterval instead

internal/integration/telemetry_sdk_emitter.go
```diff
@@ -56,10 +57,8 @@ func TelemetryHarvesterWithMetricsURL(url string) TelemetryHarvesterOpt {
 }
 
 // TelemetryHarvesterWithHarvestPeriod sets harvest period.
-func TelemetryHarvesterWithHarvestPeriod(t time.Duration) TelemetryHarvesterOpt {
-	return func(config *telemetry.Config) {
-		config.HarvestPeriod = t
-	}
+func telemetryHarvesterZeroPeriod(config *telemetry.Config) {
+	config.HarvestPeriod = 0
```
paologallinaharbur, line 61: I guess we are setting it to zero to disable the periodic harvest, did we check the behaiour? I am worried that if that is zero a different default value is used

internal/cmd/scraper/scraper.go
```diff
@@ -301,6 +300,11 @@ func Run(cfg *Config) error {
 				HarvesterOpts:                 harvesterOpts,
 				DeltaExpirationAge:            cfg.TelemetryEmitterDeltaExpirationAge,
 				DeltaExpirationCheckInternval: cfg.TelemetryEmitterDeltaExpirationCheckInterval,
+				BoundedHarvesterCfg: integration.BoundedHarvesterCfg{
+					HarvestPeriod:     hTime,
+					MinReportInterval: hTime / 10,
```
paologallinaharbur, line 305: should we set this with a hardcoded value to "save" the backend? Es 50ms?

internal/integration/bounded_harvester.go
paologallinaharbur: I guess we would need to wait anyway to increment 	h.reportedMetrics += newMetrics

internal/cmd/scraper/scraper.go
roobre: Yes, the current `/10` is very temporary. I have yet to do some load testing to find out what is a "reasonable" default. 100ms worked for our 1000 targets, but might need further refinement.

In any case, the worst case scenario for this being set "too high" is just that memory will start to increase. I wouldn't concern too much if 3000 targets caused ~1-2GiB of ram to be used, still seems reasonable to me.

internal/integration/bounded_harvester.go
```diff
@@ -0,0 +1,114 @@
...
+	defer h.mtx.Unlock()
+
+	h.reportedMetrics += newMetrics
+	if force ||
+		time.Since(h.lastReport) >= h.HarvestPeriod ||
+		(h.reportedMetrics > h.MetricCap && time.Since(h.lastReport) > h.MinReportInterval) {
+
+		h.lastReport = time.Now()
+		h.reportedMetrics = 0
+		h.inner.HarvestNow(ctx)
```
paologallinaharbur, line 98: While sending metrics we will be not able to fetch data since h.inner.HarvestNow(ctx) is in the mux, no idea if it can create issues.
The load test will solve our doubts

internal/integration/telemetry_sdk_emitter.go
roobre: Yes, it is not documented but the implementation seems clear. A period of 0 disables periodic harvesting:
https://github.com/newrelic/newrelic-telemetry-sdk-go/blob/9ff9c3035f5a4745f2f3df7ed1b13c6595542b40/telemetry/harvester.go#L96

internal/integration/bounded_harvester.go
roobre: I have also thought about this. Right now it is all done in the same function in order to keep the concurrency model simple, although it might not be the most performant option, specially if one of the calls ends up triggering a `HarvestNow`.

I'll give this a thought.

internal/integration/bounded_harvester.go
roobre: Adjusted, thx

internal/integration/bounded_harvester.go
roobre: Agreed, I have also commented this [in the other thread](https://github.com/newrelic/nri-prometheus/pull/118#discussion_r542361562). I'll give it a thought to see if we can leave the harvest process out of the critical section for checking.

I think `telemetry.Harvester` also implements some clever locking so we should not waste it on our side :)

internal/integration/bounded_harvester.go
roobre: This is now solved, `RecordMetric` no longer calls `reportIfNeeded`. Instead, the periodic routine is tasked with doing so.

This has its own pros and cons, perhaps being the most relevant one that the routine is now required for the boundedHarvester to function.

internal/cmd/scraper/scraper.go
roobre: All these parameters are now settable through the config.

internal/integration/bounded_harvester.go
roobre: Inner harvested is now longer covered inside the critical section of boundedHarvester. In addition, it takes place on an entirely different goroutine. This allows parallelizing the metric reporting.

internal/integration/telemetry_sdk_emitter.go
```diff
@@ -42,6 +42,10 @@ type TelemetryEmitterConfig struct {
 	// DeltaExpirationCheckInternval sets the cumulative DeltaCalculator
 	// duration between checking for expirations. Defaults to 30s.
 	DeltaExpirationCheckInternval time.Duration
+
+	// boundedHarvester configuration
+	DisableBoundedHarvester bool
```
paologallinaharbur, line 47: is this for future work?
I do not see where it can be configured, currently is is "hardcoded" to false?

internal/integration/bounded_harvester.go
```diff
@@ -0,0 +1,166 @@
...
+	MinReportInterval time.Duration
+
+	// DisablePeriodicReporting prevents bindHarvester from spawning the periodic report routine.
+	// It also causes an already spawned reporting routine to be stopped on the next interval.
+	DisablePeriodicReporting bool
+}
+
+// BoundedHarvesterDefaultHarvestPeriod is the default harvest period. Since harvests are also triggered by stacking
+// metrics, there is no need for this to be very low
+const BoundedHarvesterDefaultHarvestPeriod = 5 * time.Second
```
paologallinaharbur, line 71: The old default was 1 second, can we keep it as it was?

internal/integration/bounded_harvester.go
```diff
@@ -0,0 +1,166 @@
...
+func (h *boundedHarvester) RecordMetric(m telemetry.Metric) {
+	h.mtx.Lock()
+	h.storedMetrics++
+	h.mtx.Unlock()
+
+	h.inner.RecordMetric(m)
+}
+
+// HarvestNow forces a new report
+func (h *boundedHarvester) HarvestNow(ctx context.Context) {
```
paologallinaharbur, line 110: do we need this function to "satisfy" an interface requirement?
Otherwise if it is not needed by the code I would get rid of it and of the force parameter

paologallinaharbur (dismissed):
I left some minor comments, but everything  looks really good!

internal/integration/telemetry_sdk_emitter.go
roobre: I put this for the integration tests for the telemetry_sdk, as since they were not thought with the BoundedHarvester in mind, they broke in unexpected ways.

That being said, I added this option before the ability to stop the periodic harvest routine was implemented, so perhaps the tests will work nice now just with that.

I'll give it a try.

internal/integration/bounded_harvester.go
roobre: My idea was that the previous default (1s) was being a bit conservative precisely to avoid piling up many metrics in memory. Since this is no longer an issue, I think a bigger default would be more eco-friendly.

Is there another reason for shorting it that I'm not seeing?

internal/integration/bounded_harvester.go
roobre: It is mainly used for tests, but also to keep compatibility with other `Harvester` implementations. It would be impossible to wrap/compound `Harvester`s without it, in fact, `boundedHarvester` uses it precisely for this purpose :)

internal/integration/bounded_harvester.go
paologallinaharbur: not that I am aware of

internal/integration/telemetry_sdk_emitter.go
roobre: It seems that `DisablePeriodicReporting` is indeed enough and `DisableBoundedHarvester` is no longer required for the integration sdk emmiter tests to pass.

However, due to the asynchronous nature of `boundedHarvester`, this flag is still useful for `BenchmarkTelemetrySDKEmitter`.

I'd suggest to keep the flag for now, at least until performant reporting is merged to the infra-sdk and we can get rid of the asynchronicity of `boundedHarvester.HarvestNow`

paologallinaharbur (approved):

PR #125, opened 2020-12-16
ci: fix deployment of kubernetes manifests, simplify pipeline slightly

### LINKED ISSUE #115: [CI/CD] The release pipeline is no longer uploading manifests to S3
Before the release was performed in Travis:
 - https://github.com/newrelic/nri-prometheus/blob/v2.1.0/.travis.yml

Then it was moved to github Actions, however the manifests are no longer uploaded.

roobre:
This PR slightly modifies the release process so deployment yamls are generated by goreleaser as a hook, rather than with `go generate` as before.

The main reason for this change is being able to leverage goreleaser's `{{.Version}}` and `{{.Major}}.{{.Minor}}` tags, which unfortunately cannot be used in to define environment variables (i.e. in the `env` section.

The deploy-yaml script gathers these versions from the command line (passed by goreleaser) ~~and also discerns whether we are doing a prerelease or not from the env var `PRERELEASE`, which is is set by github actions.~~

Manifests are uploaded to s3 on pre-release event, like it was the case before.

### Tasks
- [x] Refactor manifest generation
- [x] Push manifests to S3

---

Fixes #115


roobre:
Rebased

roobre:
All the logic for releasing has been moved to pre-releases (like in [nri-redis](https://github.com/newrelic/nri-redis)). This is a side effect of how the pipeline was initially designed: A pre-release is done, everything gets compiled and uploaded by goreleases, then it is promoted to a release. Refactoring this logic and incorporating docker images and s3 manifests to it is very costly due to how goreleaser works, so it has been outscoped from this PR.

paologallinaharbur (approved):
Awesome, thanks, the main purpose is to fix the bug.

This whole pipeline should be refactored in the future

.github/workflows/prerelease.yml
```diff
@@ -3,10 +3,8 @@ name: Release pipeline
 on:
   release:
     types:
-       # a pre-release is created
+      # a pre-release is created
       - prereleased
-      # a release or draft of a release is published, or a pre-release is changed to a release
-      - released
     tags:
```
paologallinaharbur, line 8: is this required? isn't run twice when a release is created?

.goreleaser.yml
```diff
@@ -107,8 +112,7 @@ dockers:
       - nri-prometheus
     image_templates:
       - 'newrelic/nri-prometheus:{{ .Major }}.{{ .Minor }}'
-    # auto mean that if version has rc-1 or similar it won't be pushed
```
paologallinaharbur, line 110: rc-1?

build/release.mk
```diff
@@ -29,9 +29,14 @@ release/deps: $(GORELEASER_BIN)
 release/build: release/deps release/clean
 ifeq ($(PRERELEASE), true)
```
paologallinaharbur, line 30: I still do not understand this... the pipeline is run for prerelease and for release but we push for pre-release only?

.github/workflows/prerelease.yml
roobre: It should be run on pre-release only, that's why I removed the `- released` line.

.goreleaser.yml
roobre: Nonsense from the goreleaser docs. Apparently it won't push the image if it detects the tag has some `rc` suffixes or something like that. Since we are not using it and I didn't want that to bite us in the ass, I stripped it.

build/release.mk
roobre: Pipeline itself is run on `pre-release`, which causes GHA to set `PRERELEASE` to true. However, the `release/build` reciped target is called from another recipe that is run in the push/pr pipeline.

I think this is done to test that packages build in all arches. Rather than just copying the `goreleaser build ...` line, it is called instead. I am not very sure why.

paologallinaharbur (dismissed):

PR #126, opened 2020-12-16
Load Tests in GHA

paologallinaharbur:
This PR adds a pipeline running automatically load test each time a PR is opened and the code is pushed to main.

The image is compiled, deployed with `Skaffold`, the load test chart is deployed with 800 targets and the results from the
prometheus output are collected and parsed with a golang help tool.

Moreover I refactored the code including in the `load-test`
folder all the tool useful for that pourpose (chart, mockexporter and the dockerfile, sample data)

The `git diff` is a big confused by the refactor since the new folder is named as the old load-test chart name

Screen Shot 2020-12-18 at 10 02 18 AM

**The test fail since it is pointing to the mock data assuming they are already in main branch**


.github/workflows/load_test.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.1'
+      - name: Installing dependencies yq and skaffold
+        run: |
+          sudo wget -O /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/3.3.0/yq_linux_amd64
+          sudo chmod +x /usr/local/bin/yq
+      - name: Setup Minikube
+        uses: manusa/actions-setup-minikube@v2.0.1
```
roobre, line 24: Could we bump this to the latest version?
```suggestion
        uses: manusa/actions-setup-minikube@v2.2.0
```

paologallinaharbur: definetely

.github/workflows/load_test.yml
```diff
@@ -0,0 +1,38 @@
...
+    types:
+      - prereleased
+  pull_request:
+    branches: [ feat/automatizeLoadTests, main ]
+
+name: Running Load Tests
+jobs:
+  load_tests:
+    name: unning Load Tests
+    runs-on: ubuntu-latest
```
roobre, line 12: `setup-minikube` seems to be quite coupled to the ubuntu version ([see readme](https://github.com/manusa/actions-setup-minikube/#setup-minikube-github-action)). I'd suggest to pin this to `20.04` (which is supported in `2.2.0`)
```suggestion
    runs-on: ubuntu-20.04
```

paologallinaharbur: 👍

load-test/README.md
```diff
@@ -1,62 +1,17 @@
-# New Relic's Prometheus Load Generator
```
paologallinaharbur, line 1: This is what I mean in the PR description, git is comparing to different READMEs

gsanchezgavier (approved):
great thanks!

load-test/laod_test.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+#Clean away old resources not useful anymore
```
roobre, line 3: I'd suggest adding `set -e` to the script, which causes the whole script to stop executing and return failure when a line fails. This can help debugging and stop potentially undesired side-effects.
```suggestion
#!/bin/bash

set -e

#Clean away old resources not useful anymore
```

load-test/laod_test.sh
```diff
@@ -0,0 +1,55 @@
...
+  # dos2unix ./load-test/load_test.results
+}
+
+#Verify the results of the tests (memory, time elapsed, total targets)
+verifyResults(){
+  # we need the loadtests flag in order to make sure that these tests are run only needed
+  go test -v -tags=loadtests ./load-test/...
+}
+
+#Not used for CI, useful for localenvironemnt only
```
roobre, line 39: Why is this not used for CI? CI `run` script seems to do the same as this.

roobre:
Other than the minor comments this feels superb, great job!

load-test/laod_test.sh
paologallinaharbur: actually yes, I refactored it and I forgot to update it!

load-test/laod_test.sh
paologallinaharbur: I'll add also this

gsanchezgavier (approved):

Issue #120, opened 2020-12-14
Update prometheus and telemetry-sdk


roobre:
Right now POMI uses quite old versions of `github.com/prometheus/common`. `github.com/prometheus/client_golang` and `github.com/newrelic/newrelic-telemetry-sdk-go`.

It would be nice to bump those versions to benefit from potential performance improvements and bug fixes.

Issue #116, opened 2020-12-11
403 response on telemetry SDK


gsanchezgavier:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )
Some request are being rejected by Telemetry API with 413 error. this packages are dropped and the metric lost. 
## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )
The packages are corectly choped by the telemetry sdk 

## [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

## Steps to Reproduce
[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )
In a high load environment (the chance to reproduce it increase) look at the log message. We start to see this error with more than 400 targets. 
## Your Environment
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Additional context
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )
A PR has been submitted to the sdk . After [this is released](https://github.com/newrelic/newrelic-telemetry-sdk-go/pull/39), the dependency should be updated.

RichVanderwal:
The [PR](https://github.com/newrelic/newrelic-telemetry-sdk-go/pull/39) has been merged, and is awaiting the next release of the Go Telemetry SDK.

paologallinaharbur:
@RichVanderwal is it scheduled already next release?

PR #127, opened 2020-12-17
feat(dependencies): updated telemetrySDK, promClient

paologallinaharbur:
Updating the dependency of telemetrySDK and of the prometheus client we solve the "entity too large 413" issue and we avoid the risk to fall too "behind" with client versions

fix https://github.com/newrelic/nri-prometheus/issues/116
fix https://github.com/newrelic/nri-prometheus/issues/120


go.mod
```diff
@@ -8,22 +8,21 @@ require (
...
 	github.com/onsi/ginkgo v1.10.1 // indirect
 	github.com/onsi/gomega v1.7.0 // indirect
 	github.com/pelletier/go-toml v1.2.1-0.20181124002727-27c6b39a135b // indirect
-	github.com/pkg/errors v0.8.1
-	github.com/prometheus/client_golang v0.9.3
-	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
-	github.com/prometheus/common v0.4.0
-	github.com/sirupsen/logrus v1.3.0
+	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.9.0
```
roobre, line 16: Just checked the breaking changes according to [the changelog](https://github.com/prometheus/client_golang/blob/master/CHANGELOG.md#100--2019-06-15) and we don't seem to be using anything there.

roobre (approved):
lgtm

go.mod
```diff
@@ -8,22 +8,21 @@ require (
 	github.com/hashicorp/hcl v1.0.1-0.20190611123218-cf7d376da96d // indirect
 	github.com/imdario/mergo v0.3.8 // indirect
 	github.com/newrelic/infra-integrations-sdk v1.0.1-0.20201016141521-6fdfe28ad6c0
```
roobre, line 10: Just a small comment tho, have we tried to upgrade this and see what breaks?

roobre: It would seem that [v3.0.0](https://github.com/newrelic/infra-integrations-sdk/blob/master/CHANGELOG.md#300-beta) onwards removed protocol v1, so let's hold this for now.

PR #135, opened 2021-01-05
dockerfile: upgrade base image to alpine:3.12

### LINKED ISSUE #134: Upgrade base image to alpine:3.12

## Description

`alpine:3.9` contains [`OpenSSL 1.1.1g`](https://pkgs.alpinelinux.org/packages?name=openssl&branch=v3.9), which has a [security vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2020-1971).

Upgrading is not trivial as the release dockerfile forces `musl=1.1.20-r5` for reasons that are not clear to me.

roobre:
Fixes #134


Dockerfile.release
```diff
@@ -1,11 +1,9 @@
-FROM alpine:3.9
+FROM alpine:3.12
 
 ARG BINARY=nri-prometheus
 
 RUN apk add --no-cache --upgrade \
-        ca-certificates \
-        musl=1.1.20-r5
```
roobre, line 7: It is unclear to me why this specific version of musl was pinned before. I did some basic sanity checks (i.e. checking that the integration launches) without it and it seems to run properly.

roobre: @carlosroman According to the blame it was you who introduced this pin, do you by chance remember the reason? Thanks!

arvdias: according to this list of packages https://pkgs.alpinelinux.org/packages?name=musl&branch=v3.12&repo=main&arch=x86_64. Alpine 3.12 has a more recent version of musl, so we should be fine removing that pinned version (unless the pinned version was for some backwards compat issue)

arvdias (approved):

Issue #136, opened 2021-01-19
Regenerate windows installer GUIDs and upgrade codes


roobre:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )

Some upgrade codes and GUIDs are identical to the ones used in https://github.com/newrelic/nri-redis, and should be changed.

PR #138, opened 2021-01-20
fix(ci/cd): build pipeline and windows package codes

paologallinaharbur:
This Pr fixes:
 - the windows package upgrade codes
 - fixes the Pr pipelines adding docker login
 - removed placing in the right scope some of the secrets
 
 
 Fix https://github.com/newrelic/nri-prometheus/issues/136

### DIFF
--- a/build/package/windows/nri-386-installer/Product.wxs
+++ b/build/package/windows/nri-386-installer/Product.wxs
@@ -2,7 +2,7 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
     <?define Version = "$(var.IntegrationVersion)" ?>
     <?define ProductId = "*" ?>
-    <?define UpgradeCode = "{e39026bc-1443-11eb-adc1-0242ac120002}" ?>
+    <?define UpgradeCode = "{aaf34f54-a732-4ba8-b85c-d073347d7e99}" ?>
 
     <Product Id="$(var.ProductId)"
              Name="New Relic Infrastructure Integration, nri-$(var.IntegrationName)"
@@ -27,8 +27,6 @@
             <ComponentRef Id="CMP_V1_CUSTOM_PLUGINS"/>
             <ComponentRef Id="CMP_V1_CUSTOM_PLUGINS_BIN"/>
             <ComponentGroupRef Id="CustomPluginsBinComponent"/>
-            <!-- <ComponentGroupRef Id="CustomPluginsComponent"/>
-            <ComponentGroupRef Id="PluginConfigsComponent"/> -->
         </Feature>
     </Product>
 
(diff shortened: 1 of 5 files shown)


build/package/windows/nri-386-installer/Product.wxs
```diff
@@ -49,51 +47,34 @@
...
-                      Source="$(var.BinariesPath)New Relic\newrelic-infra\newrelic-integrations\$(var.IntegrationName)-win-definition.yml"
-                      KeyPath="yes"/>
-            </Component>
-        </ComponentGroup>
-        <ComponentGroup Id="PluginConfigsComponent" Directory="PluginConfigsFolder">
-            <Component Id="CMP_NRI_$(var.IntegrationName)_CONFIG_YML" Guid="829cd54e-8454-44c6-aec6-9e6b38373717" Win64="no"
-                       NeverOverwrite="yes" Permanent="yes">
-                <File Id="FILE_NRI_$(var.IntegrationName)_CONFIG_YML"
-                      Name="$(var.IntegrationName)-win-config.yml.sample"
-                      Source="$(var.BinariesPath)New Relic\newrelic-infra\integrations.d\$(var.IntegrationName)-win-config.yml.sample"
```
gsanchezgavier, line 94: looks this is not the [config](https://github.com/newrelic/nri-prometheus/blob/main/configs/nri-prometheus-config.yml.sample) in the repo . Why we are not adding configs to msi?

build/package/windows/nri-amd64-installer/Product.wxs
```diff
@@ -49,51 +47,34 @@
...
-                      Source="$(var.BinariesPath)New Relic\newrelic-infra\newrelic-integrations\$(var.IntegrationName)-win-definition.yml"
-                      KeyPath="yes"/>
-            </Component>
-        </ComponentGroup>
-        <ComponentGroup Id="PluginConfigsComponent" Directory="PluginConfigsFolder">
-            <Component Id="CMP_NRI_$(var.IntegrationName)_CONFIG_YML" Guid="1bf72f1b-9d68-4c59-bd52-be6b7fe2e107" Win64="yes"
-                       NeverOverwrite="yes" Permanent="yes">
-                <File Id="FILE_NRI_$(var.IntegrationName)_CONFIG_YML"
-                      Name="$(var.IntegrationName)-win-config.yml.sample"
-                      Source="$(var.BinariesPath)New Relic\newrelic-infra\integrations.d\$(var.IntegrationName)-win-config.yml.sample"
```
gsanchezgavier, line 94: same here

build/package/windows/nri-386-installer/Product.wxs
paologallinaharbur: That was only a comment that I removed. 

I believe that we are not adding it since it is going to be generated and the user should not configure it.
Currently Linux behaves in the same way (no config files pre-packaged), therefore I believe it was a decision and not a mistake.

@arvdias @ardias Do you remember why we were not packaging configs?

gsanchezgavier: yes you are right , this is bundle with the agent. We can include it later if needed

gsanchezgavier (approved):

PR #140, opened 2021-01-22
Drop named attributes by metric prefix

stevenosborne-wf:
We have some prometheus metrics we forward to New Relic that have very high cardinality due to attributes we don't need or use. Given that for many prometheus exporters, we can not configure this, we would like to drop these attributes at the scraper.


paologallinaharbur:
Stale PR, closing.

PR #142, opened 2021-02-10
Add configuration sample to packages

skatsaounis:
This PR adds configuration sample to nix packages. Furthermore, it removes nri-* prefix to conform to documentation examples of integrations' configuration files.


gsanchezgavier:
@skatsaounis Thanks for the contribution !, It would be really valuable if you could describe your use case for the prometheus integration, like where you are installing it and what kind of targets are you scraping. Thanks!

skatsaounis:
Hi @gsanchezgavier. To be honest, I was trying to install the newrelic-infra on a virtual machine and configure it appropriately to collect Hashicorp Vault telemetry data. During this experiment with the tool, I struggled to find an example in documentation since everything was referring to Kubernetes. After some investigation, I found what I had to do.

In the meantime, I had no problem to collect Consul metrics with nri-consul. I noticed that nri-consul .deb package contains the configuration sample https://github.com/newrelic/nri-consul/blob/master/build/.goreleaser.yml#L49. With this PR I aim to make it easier for future users to configure nri-prometheus outside Kubernetes deployments.

PS. I have already opened this PR as well in nri-consul (https://github.com/newrelic/nri-consul/pull/43) and I am preparing a new one to switch to the new integration configuration which is present in `consul-config.yml.k8s_sample` but not in the plain `consul-config.yml.sample`.

paologallinaharbur:
Stale PR, closing for now

PR #143, opened 2021-02-17
multi arch support for docker image

paologallinaharbur:
This PR adds support for arm64 leveraging goreleaser to create the manifest

Moreover it adds tini as entrypoint


.goreleaser.yml
```diff
@@ -104,25 +104,39 @@ release:
...
-      - nri-prometheus
+    ids:
+      - nri-prometheus-nix
     image_templates:
-      - 'newrelic/nri-prometheus:{{ .Version }}'
+      - 'pgallina/nri-prometheus-test:{{ .Version }}-arm64'
+    use_buildx: true
+    build_flag_templates:
+    - "--platform=linux/arm64"
     skip_push: false
```
roobre, line 128: Could we add `arm` as well here? I think we're publishing other integrations and the agent for arm(32).

Dockerfile.release
```diff
@@ -1,15 +1,19 @@
 FROM alpine:3.12
```
roobre, line 1: Can be bumped to 3.13 :)

Dockerfile.release
```diff
@@ -1,15 +1,19 @@
...
+ARG TINI_VERSION=v0.19.0
 
 RUN apk add --no-cache --upgrade \
-        ca-certificates
+        ca-certificates \
+        curl
 
-COPY ${BINARY} /bin/nri-prometheus
+RUN curl -fsSLo /sbin/tini "https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-static-${TARGETARCH}" && \
+    chmod +x /sbin/tini
```
roobre, line 11: I'd rather `apk add tini` instead.

Dockerfile.release
```diff
@@ -1,15 +1,19 @@
...
 
+COPY ./nri-prometheus /bin/nri-prometheus
 USER nobody
 
 # When standalone is set to true nri-prometheus does not require an infrastructure agent to work and send data
 ENV STANDALONE=TRUE
 
-ENTRYPOINT ["/bin/nri-prometheus"]
+ENTRYPOINT ["/sbin/tini", "--", "/bin/nri-prometheus"]
```
roobre, line 19: POSIX hell awaits to those who dare miss the newline before EOF

Dockerfile.release
paologallinaharbur: I am doing this since I needed the static tini version, when installing the classical one it was failing due to missing DDLS in arm64

Dockerfile.release
paologallinaharbur: sure

Dockerfile.release
paologallinaharbur: I'll test it again, it it possible that it was failing since it was fetching the wrong image at first

roobre (approved):
LGTM.
Just a reminder that we need to change temporary creds/uncomment s3 stuff before merging.

davidbrota (dismissed):

Dockerfile.release
```diff
@@ -1,15 +1,15 @@
-FROM alpine:3.12
-
-ARG BINARY=nri-prometheus
+FROM alpine:3.13
+ARG TARGETOS
+ARG TARGETARCH
```
roobre, line 3: Are we actually using these?

paologallinaharbur: we were, it is legacy from the first commit of this PR. 🤣 

I'll get rid of it

roobre (approved):
LGTM

roobre:
Just saw we use `GO_VERSION: '1.15.2'`. Perhaps we can make a quick change and upgrade to `1.16`?

paologallinaharbur:
before merging change back commented out code and `pgallina/` repository for images

roobre (approved):
LGTM

PR #147, opened 2021-03-09
chore: remove definition path from config sample

jlegoff:
These definitions aren't ready to be used - it's better to remove it
from the config sample.


davidbrota:
there are more config files using that config option in the deploy folder. Could you please remove them too?

Issue #27, opened 2019-10-29
Services are scraped directly instead of service endpoints


nlowe:
While services with the a `prometheus.io/scrape` annotation can be discovered, it appears that `nri-prometheus` scrapes the service itself and not service endpoints:

https://github.com/newrelic/nri-prometheus/blob/d8a8f1b1c4ff34edc8d78ceecb422cc63b29e163/internal/pkg/endpoints/kubernetes.go#L178-L192

Usually you want the [service endpoints](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#endpoints) to be discovered instead. This is useful for applications like kube/coredns that only expose the application port (`53`) on the service and not the metrics port (`9153`):

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    prometheus.io/port: "9153"
    prometheus.io/scrape: "true"
  labels:
    eks.amazonaws.com/component: kube-dns
    k8s-app: kube-dns
    kubernetes.io/cluster-service: "true"
    kubernetes.io/name: CoreDNS
  name: kube-dns
  namespace: kube-system
#...
```

In the current configuration, `nri-prometheus` cannot scrape metrics from kube-dns because the port indicated by `prometheus.io/port` isn't exposed on the service:

```txt
default/nri-prometheus-f99cfb76b-h7hpw[nri-prometheus]: time="2019-10-29T20:26:41Z" level=warning msg="fetching Prometheus: http://kube-dns.kube-system.svc:9153/metrics (kube-dns)" component=Fetcher error="Get http://kube-dns.kube-system.svc:9153/metrics: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"
```

This also presents a problem for applications that ***do*** expose metrics on the same port as the service: `nri-prometheus` won't have a complete view of the application because it'll only scrape a single pod each scrape interval instead of all pods that back the service.

I could manually update the target list every time a new pod gets deployed but that's not exactly ideal.

douglascamata:
Hi @nlowe, thanks for the report. 

Let me first quickly explain that we have 2 use cases while scraping services:

1. Where a single endpoint behind the service can provide all the metrics;
2. Where all the endpoints behind the service need to scraped for metric collection to be considered complete.

Your proposal fixes point 2 while breaking point 1 though. 

> I could manually update the target list every time a new pod gets deployed but that's not exactly ideal.

[...]

PR #148, opened 2021-03-09
 fix(endpoints): scrape endpoints and not services

paologallinaharbur:
Fix https://github.com/newrelic/nri-prometheus/issues/27

The idea is that when a service is detected the underlying endpoints are fetched. For backward compatibility the metric directly coming from the service is kept
Screen Shot 2021-03-10 at 11 57 09 AM

Two new config options have been added `ScrapeServices (default true)` and `ScrapeEndpoints(default false)`
 
 After merging we should update documentation and the helm chart to ask for listing/watching permission on endpoints

### DIFF
--- a/cmd/k8s-target-retriever/main.go
+++ b/cmd/k8s-target-retriever/main.go
@@ -33,7 +33,7 @@ func main() {
 	}
 
 	kubeconf := endpoints.WithKubeConfig(*kubeConfigFile)
-	ktr, err := endpoints.NewKubernetesTargetRetriever("prometheus.io/scrape", false, kubeconf)
+	ktr, err := endpoints.NewKubernetesTargetRetriever("prometheus.io/scrape", false, true, true, kubeconf)
 	if err != nil {
 		logrus.Fatalf("could not create KubernetesTargetRetriever: %v", err)
 	}
--- a/cmd/k8s-target-retriever/main.go
+++ b/cmd/k8s-target-retriever/main.go
@@ -55,7 +55,5 @@ func main() {
 			logrus.Infof("%s[%s] %s", b.Name, b.Object.Kind, b.URL.String())
 		}
 		logrus.Infof("###################################")
-
-		logrus.Println()
 	}
 }
--- a/cmd/k8s-target-retriever/main.go
+++ b/cmd/k8s-target-retriever/main.go
@@ -44,14 +44,17 @@ func main() {
 
 	logrus.Infoln("connected to cluster, watching for targets")
 
-	for range time.Tick(time.Second * 5) {
+	for range time.Tick(time.Second * 7) {
 		targets, err := ktr.GetTargets()
+		logrus.Infof("###################################")
+
 		if err != nil {
 			logrus.Fatalf("could not get targets: %v", err)
 		}
 		for _, b := range targets {
 			logrus.Infof("%s[%s] %s", b.Name, b.Object.Kind, b.URL.String())
 		}
+		logrus.Infof("###################################")
 
 		logrus.Println()
 	}
--- a/cmd/nri-prometheus/config.go
+++ b/cmd/nri-prometheus/config.go
@@ -109,6 +109,8 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("insecure_skip_verify", false)
 	viper.SetDefault("standalone", true)
 	viper.SetDefault("disable_autodiscovery", false)
+	viper.SetDefault("scrape_services", true)
+	viper.SetDefault("scrape_endpoints", false)
 	viper.SetDefault("percentiles", []float64{50.0, 95.0, 99.0})
 	viper.SetDefault("worker_threads", 4)
 }
--- a/deploy/nri-prometheus.tmpl.yaml
+++ b/deploy/nri-prometheus.tmpl.yaml
@@ -18,6 +18,7 @@ rules:
     - "nodes/proxy"
     - "pods"
     - "services"
+    - "endpoints"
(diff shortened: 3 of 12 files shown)


paologallinaharbur:
Currently after merging this we will be sending more data than before.

We should decide to filter out some targets, es:
```go
func (k *KubernetesTargetRetriever) GetTargets() ([]Target, error) {
	length := 0
	k.targets.Range(func(_, _ interface{}) bool {
		length++
		return true
	})
	targets := make([]Target, 0, length)
	k.targets.Range(func(_, y interface{}) bool {
		for _, t := range y.([]Target) {
			if t.Object.Kind == "service" && excludeServiceTargets {
				continue
			}
			if t.Object.Kind == "endpoints" && excludeEndpointsTargets {
				continue
			}
			targets = append(targets, t)
		}
		return true
	})

	return targets, nil
}
```

internal/pkg/endpoints/kubernetes.go
```diff
@@ -511,19 +598,24 @@ func (k *KubernetesTargetRetriever) processEvent(event watch.Event, requireLabel
...
 				k.addTarget(object, event.Type)
 				return
 			}
 			// If the object is not scrapable and we've seen it before, we remove it.
 			if !scrapable && seen {
 				k.targets.Delete(string(object.GetUID()))
 				debugLogEvent(klog, event.Type, "deleted", object)
+				switch obj := object.(type) {
+				case *apiv1.Service:
+					if e, err := k.client.CoreV1().Endpoints(obj.Namespace).Get(obj.Name, metav1.GetOptions{}); err == nil {
```
gsanchezgavier, line 612: we might need to cache the services and the endpoints related to avoid the calls to the API

gsanchezgavier (commented):
I think we need  to have these behind a feature flag keeping the old behavior by default. Or have a mayor release. I would also propose to have this feature in beta for some time.  

Could you try the load test to se how is the result with the new feature?

cmd/k8s-target-retriever/main.go
```diff
@@ -44,15 +44,16 @@ func main() {
 
 	logrus.Infoln("connected to cluster, watching for targets")
 
-	for range time.Tick(time.Second * 5) {
+	for range time.Tick(time.Second * 7) {
```
davidbrota, line 47: why 7 and not 5? should it be made configurable?

paologallinaharbur: this is in the "test" main, it is not used by the real pomi

paologallinaharbur: it is used to test an debug the targets retrieved

paologallinaharbur:
@gsanchezgavier I added two feature flags and given what we talked about offline in the past days, for now the option to scrape endpoints will be disabled by default.

I'll update the documentation and the HelmChart to further clarify this

gsanchezgavier (approved):
lgtm

paologallinaharbur:
After merging and releasing we need to update documentation and expose the new config through the helm charts

cmd/nri-prometheus/config.go
```diff
@@ -109,6 +109,8 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("insecure_skip_verify", false)
 	viper.SetDefault("standalone", true)
 	viper.SetDefault("disable_autodiscovery", false)
+	viper.SetDefault("enable_scrape_services", true)
+	viper.SetDefault("enable_scrape_endpoints", false)
```
roobre, line 113: I'd rename these without the `enable_` prefix

```suggestion
	viper.SetDefault("scrape_services", true)
	viper.SetDefault("scrape_endpoints", false)
```

deploy/nri-prometheus.tmpl.yaml
```diff
@@ -114,11 +115,19 @@ data:
     # The label used to identify scrapable targets. Defaults to "prometheus.io/scrape".
     scrape_enabled_label: "prometheus.io/scrape"
 
+    # enable_scrape_services Allows to enable scraping the service and not the endpoints behind.
+    # When endpoints are scraped this is no longer needed
```
roobre, line 119: Could we check this in the code? Since it is true by default, it is possible that some users set `enable_scrape_endpoints: true` but forget to disable the old one.

In fact, if the options are mutually exclusive, I'd strip `enable_scrape_services` entirely.

paologallinaharbur: I am not totally sure, but I can imagine a possible use-case for which they actually want both.
Otherwise it would be quite obscure that enabling endpoints you lose services

cmd/nri-prometheus/config.go
paologallinaharbur: mmm, I added it since there was just above disale_... but I can change that :)

internal/pkg/endpoints/kubernetes.go
```diff
@@ -511,19 +610,24 @@ func (k *KubernetesTargetRetriever) processEvent(event watch.Event, requireLabel
...
+						k.targets.Delete(string(e.GetUID()))
+						debugLogEvent(klog, event.Type, "deleted", e)
+					}
+				}
 			}
-		}
-		if !requireLabel {
+		} else if !requireLabel {
```
roobre, line 630:
```suggestion
		} else {
```

paologallinaharbur: didn't you saw "Please, do not try to reduce the amount of code below or simplify the conditionals."?

I'll update that xD

roobre (approved):

PR #149, opened 2021-03-17
do not scrape pods that are completed

anandkumarpatel:
Pods that are completed and will never be restarted should not be scrapped because there is no process alive to handle the call. 

This PR add a check for pod completion to the `isObjectScrapable` function. 

I made it a switch statement in case there is a need to use object-specific keys to determine scrapability in the future. 

[Pod Phases doc](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase)
[Restart Policy doc](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy)

Fix for: https://github.com/newrelic/nri-prometheus/issues/150

### DIFF
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -155,6 +155,13 @@ func (k *KubernetesTargetRetriever) listServices() error {
 }
 
 func isObjectScrapable(o metav1.Object, label string) bool {
+	switch obj := o.(type) {
+	case *apiv1.Pod:
+		if isPodCompleted(obj) {
+			return false
+		}
+	}
+
 	return o.GetLabels()[label] == trueStr || o.GetAnnotations()[label] == trueStr
 }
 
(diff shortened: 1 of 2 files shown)


paologallinaharbur:
We are currently modifying the same snippet of the code you are modifying.
Therefore likely we will need to rebase this PR once the one enabling endpoints gets merged.
By the way thanks for the contribution!

anandkumarpatel:
Hey @paologallinaharbur I am fine with rebasing. I took a look at the other PR and they are compatible. I only modify `isObjectScrapable` and look like the other PR does not modify that function. 

But would it be possible to merge this one first? Seems like the other one might be a breaking change. I'll be happy to rebase into the other branch as well to make it possible!

paologallinaharbur:
Closing since there has been no activity. 
Feel free to propose a new one or reopen this one :)

anandkumarpatel:
@paologallinaharbur Created new one here: https://github.com/newrelic/nri-prometheus/pull/212

PR #155, opened 2021-04-19
log deltas being dropped due to invalid calculation

roobre:
This PR adds an INFO logs when datapoints of DELTA (COUNTER) metrics cannot be calculated, which may happen when such points do not arrive in order from the exporter.

### DIFF
--- a/internal/integration/telemetry_sdk_emitter.go
+++ b/internal/integration/telemetry_sdk_emitter.go
@@ -207,9 +207,11 @@ func (te *TelemetryEmitter) Emit(metrics []Metric) error {
 				metric.value.(float64),
 				now,
 			)
-			if ok {
-				te.harvester.RecordMetric(m)
+			if !ok {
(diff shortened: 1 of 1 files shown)


internal/integration/telemetry_sdk_emitter.go
```diff
@@ -207,9 +207,11 @@ func (te *TelemetryEmitter) Emit(metrics []Metric) error {
 				metric.value.(float64),
 				now,
 			)
-			if ok {
-				te.harvester.RecordMetric(m)
+			if !ok {
+				logrus.Infof("Got invalid delta for %s=%v, skipping", metric.name, metric.value)
```
paologallinaharbur, line 211: Can we change this to be debug? I am only worried this can fload us depending on the # of times it happens.
Do we have any idea?

roobre:
Closing for now as this has become stale with no further interest in getting this logged.

PR #156, opened 2021-04-21
Buffer the full page in the scraper to expose unexpected EOF.

a-feld:
Hi there folks! 👋 

We're seeing some pretty wild flapping of our SLI which is a composition of 2 different metrics. We see this flapping on deploy.

We believe we've tracked this down to the behavior of nri-prometheus when a page is only partially loaded:
https://github.com/a-feld/prometheus-unexpected-eof

When a page is partially loaded, an `unexpected EOF` error is raised, but is squashed by the prometheus decoder. This PR exposes the `unexpected EOF` error (io.ErrUnexpectedEOF) and avoids only partially reporting metrics which can lead to weird behavior of metrics that are compositions in the UI.

### DIFF
--- a/internal/pkg/prometheus/prometheus.go
+++ b/internal/pkg/prometheus/prometheus.go
@@ -57,9 +57,9 @@ func Get(client HTTPDoer, url string) (MetricFamiliesByName, error) {
 	if err != nil {
 		return mfs, err
 	}
-	b := bytes.NewBuffer(body)
+	r := bytes.NewReader(body)
 
-	d := expfmt.NewDecoder(b, expfmt.FmtText)
+	d := expfmt.NewDecoder(r, expfmt.FmtText)
 	for {
 		var mf dto.MetricFamily
 		if err := d.Decode(&mf); err != nil {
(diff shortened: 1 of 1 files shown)


a-feld:
jenkins retest this please

internal/pkg/prometheus/prometheus.go
```diff
@@ -62,16 +49,17 @@ func Get(client HTTPDoer, url string) (MetricFamiliesByName, error) {
...
 		return nil, fmt.Errorf("status code returned by the prometheus exporter indicates an error occurred: %d", resp.StatusCode)
 	}
 
-	countedBody := &countReadCloser{innerReadCloser: resp.Body}
-	d := expfmt.NewDecoder(countedBody, expfmt.FmtText)
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return mfs, err
+	}
+	b := bytes.NewBuffer(body)
```
roobre, line 60: Perhaps we could use `bytes.NewReader` rather than `NewBuffer`, which afaik it's cheaper :)

a-feld: ooh I'll give that a shot! 💡

roobre:
Merging to an internal branch for CI to do it's thing

PR #175, opened 2021-05-10
Update dependencies

nadiamoe:


roobre:
@paologallinaharbur Many were indirect dependencies that were no longer needed since (presumably) the direct dependencies switched to go modules already :)

PR #182, opened 2021-05-26
endpoints: check for path prefix with strings.HasPrefix

### LINKED ISSUE #181: getPath panics attempting to index an empty label value
## Description

The `getPath` function checks whether a series of labels are present in a Kubernetes objects, and disregard the value if the label is not found. However, labels are only checked for existence, not for value emptyness:

https://github.com/newrelic/nri-prometheus/blob/3a8cceb79981cc083e4a1a16eadcf6618230d954/internal/pkg/endpoints/kubernetes.go#L292-L298

A few lines later, the label is indexed with `[0]`. This will cause a panic if the label is empty.

https://github.com/newrelic/nri-prometheus/blob/3a8cceb79981cc083e4a1a16eadcf6618230d954/internal/pkg/endpoints/kubernetes.go#L300

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )

* ~~`nri-prometheus` should ignore empty lables, just like it ignores non-existent ones.~~
* `nri-prometheus` should handle empty lables, assuming `/`.

roobre:
Previously, path[0] was used which caused a panic for empty labels
This will _allow_ the use of an empty label value, which will be translated to "/"

Fixes #181

### DIFF
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -286,20 +286,20 @@ func getPortList(e *apiv1.Endpoints, s *apiv1.Service) []string {
 
 func getPath(o metav1.Object) string {
 	var path string
-	var ok bool
 
 	// Annotations take precedence over labels.
-	if _, ok = o.GetAnnotations()[defaultScrapePathLabel]; ok {
-		path, _ = o.GetAnnotations()[defaultScrapePathLabel]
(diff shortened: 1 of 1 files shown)


paologallinaharbur (approved):
only check if there are similar functions in the code, because I remember seeing that more than once :)

roobre:
@paologallinaharbur I found a second place where slices were indexed without length check, but it was safe since the source of the data was controlled.

Merging!

PR #189, opened 2021-06-03
internal/pkg/endpoints: run unit tests in parallel

invidian:
That cuts down test execution time by half (from 2.7s to 1.3s in my
case).

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>


roobre (approved):
👍🏻

PR #192, opened 2021-06-04
ci: move vuln checkers to a different workflow

roobre:


.github/workflows/security.yml
```diff
@@ -0,0 +1,34 @@
+name: Security scan
+on:
+  push:
+    branches:
+      - master
+      - main
+  pull_request:
+  schedule:
+    - cron: "0 3 * * MON" # Every monday at 3 AM
```
gsanchezgavier, line 9: how we do get notified on this failing ?

gsanchezgavier (approved):
lgtm

.github/workflows/security.yml
roobre: Right now we do not, unless we set up some failing CI notifications on GH. We do get notified when it fails on a synchronous action (e.g. a PR) so it's the same functionality as before, but paving the ground for notifications later.

We would need a dedicated channel for handling these notifications. I think.

PR #194, opened 2021-06-04
internal: various improvements

invidian:
This PR brings various improvements to `internal/pkg/endpoints` and `internal/integration` packages like better formatting, imports, parallel tests. There is no functional changes, but I believe those changes should make code easier to read and understand.

There are some changes to exported types to simplify the APIs of internal packages.


roobre (approved):
lgtm

PR #197, opened 2021-06-09
Improve formatting in packages and make imports more consistent

invidian:


internal/integration/fetcher.go
```diff
@@ -14,7 +14,7 @@ import (
 	"sync"
 	"time"
 
-	io_prometheus_client "github.com/prometheus/client_model/go"
+	dto "github.com/prometheus/client_model/go"
```
roobre, line 17: I'm not really a fan of the d-word... Could we use `model` instead?

invidian: This is how it is imported in Prometheus code itself: https://pkg.go.dev/github.com/prometheus/common/expfmt. I think we should be following that.

invidian:
Rebased to resolve conflicts.

roobre (approved):

invidian:
CI blocked on #221.

PR #207, opened 2021-06-23
feat: Add entity synthesis

gsanchezgavier:
Fixes https://github.com/newrelic/newrelic-coreint/issues/108


internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,172 @@
...
+			er.EntityType = ed.EntityType
+			er.tagRules = er.Tags.getTagRules()
+			er.tagRules = append(er.tagRules, ed.EntityRule.tagRules...)
+			s.addConditions(er)
+		}
+	}
+	return s
+}
+func (s *Synthesizer) addConditions(rule EntityRule) {
+	if rule.Identifier == "" || rule.Name == "" || rule.EntityType == "" {
```
ivancorrales, line 38: Just  thinking if defining a method in EntityRule would make easy to understand this  validation

```go
func (rule *EntityRule) isValid() bool{
	return  rule.Identifier != "" && rule.Name != "" && rule.EntityType != "" 
}
```
Don't know if ìsVali would be the correct name for the function or maybe other name suits better but what I mean is this piece of code only belongs to the EntityRule and the type should provide a method for check it

```go
unc (s *Synthesizer) addConditions(rule EntityRule) {
	if !rule.isValid() {
...
}
```

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Condition contains parameters used to match entities from metrics
+type Condition struct {
+	Attribute string `mapstructure:"attribute"`
+	Prefix    string `mapstructure:"prefix"`
+	Value     string `mapstructure:"value"`
+}
+
+func (c Condition) match(attribute string) bool {
```
ivancorrales, line 69: I guess we could remove the second ìf`. because if c.Prefix==""  the has.Prefix will return true always.

```go
func (c Condition) match(attribute string) bool {
	if c.Value != "" {
		return c.Value == attribute
	}
	if c.Prefix != "" {
		return strings.HasPrefix(attribute, c.Prefix)
	}
	// if Value and Prefix are empty there is a match since the attribute exists
	return true
}

func (c Condition) match2(attribute string) bool {
	if c.Value != "" {
		return c.Value == attribute
	}
	return strings.HasPrefix(attribute, c.Prefix)
}
```

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,172 @@
...
+// based on the metric attributes.
+func (s Synthesizer) GetEntityMetadata(m Metric) (sdk_metadata.Metadata, bool) {
+	rule, found := s.getMatchingRule(m)
+	if !found {
+		return sdk_metadata.Metadata{}, false
+	}
+
+	entityName := getEntityAttribute(m.attributes, rule.Identifier)
+	entityDisplayName := getEntityAttribute(m.attributes, rule.Name)
+	if entityName == "" || entityDisplayName == "" {
```
ivancorrales, line 113: NIT: If enityName=="" , we are doing an unnecessary  call to find the entityDisplayName

The other alternative to this, it would be if-if ... 
```go
if entityName := getEntityAttribute(m.attributes, rule.Identifier);entityName!=""{
		if entityDisplayName := getEntityAttribute(m.attributes, rule.Name); entityDisplayName!="" {
			md := sdk_metadata.New(rule.EntityType + ":" + entityName, rule.EntityType, entityDisplayName)
			// Adds attributes as entity tag, sdk adds the prefix "tags." to the key.
			for _, t := range rule.tagRules {
				if tagVal, ok := m.attributes[t.name]; ok {
					md.AddTag(t.entityTagName, tagVal)
				}
			}

			return *md, true
		}
	}

	return sdk_metadata.Metadata{}, false
```

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,172 @@
...
+	if len(matches) == 0 {
+		return
+	}
+	if len(matches) > 0 {
+		// longer prefix matches take precedences over shorter ones.
+		// this allows to discriminate "foo_bar_" from "foo_" kind of metrics.
+		sort.Slice(matches, func(i, j int) bool { return len(matches[i].Prefix) > len(matches[j].Prefix) })
+	}
+	rule, found = s.rulesByConditions[matches[0]]
+	return
```
ivancorrales, line 158: This return could be removed

ivancorrales, line 158: Actually,  the below code should be enough

```go
if len(matches) > 0 {
		// longer prefix matches take precedences over shorter ones.
		// this allows to discriminate "foo_bar_" from "foo_" kind of metrics.
		sort.Slice(matches, func(i, j int) bool { return len(matches[i].Prefix) > len(matches[j].Prefix) })
		rule, found = s.rulesByConditions[matches[0]]
	}
```

ivancorrales (commented):
Just a couple of comments but it looks good to me

internal/integration/entity_synthesis.go
gsanchezgavier: I think it will be possible , and i will leave the same comment to clarify the intention behind the logic

internal/integration/entity_synthesis.go
gsanchezgavier: is true there will be an unnecessary call , but since the cost of that is really small i keep it simple like this to favor readability. I think it will be harder to read if we go for the nested approach here without too much gain on performance.

internal/integration/entity_synthesis.go
gsanchezgavier: this will cause that unique matches rules will return empty rule.
We need `rule, found = s.rulesByConditions[matches[0]]` to cover that case

gsanchezgavier: this cannot be implicit

internal/integration/entity_synthesis.go
gsanchezgavier: yes now is more readable , thanks

ivancorrales (approved):

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+			er.EntityType = ed.EntityType
+			er.tagRules = append(er.Tags.getTagRules(), ed.EntityRule.tagRules...)
+			s.addConditions(er)
+		}
+	}
+	return s
+}
+func (s *Synthesizer) addConditions(rule EntityRule) {
+	if !rule.isValid() {
+		return
```
roobre, line 37: I'd return an `error` here, we do not want to silently ignore malformed rules. The constructor could also return `(Synthesizer, error)` and have this error checked when the `Synthesizer` is created, acting upon it (likely `log.Error`-ing it).

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+import (
+	"strings"
+
+	sdk_metadata "github.com/newrelic/infra-integrations-sdk/v4/data/metadata"
+	"github.com/newrelic/nri-prometheus/internal/pkg/labels"
+)
+
+// Synthesizer group of rules to synthesis entities
+type Synthesizer struct {
+	rulesByConditions map[Condition]EntityRule
```
roobre, line 15: It would seem that this map is used as a way to keep a list of `Condition`s and the `Rule`s the would apply if matched. This makes sense but I wonder if, rather than a map, we could use a list instead:
```
[]struct{
  cond Condition
  rule Rule
}
```
Lists are cheaper to traverse, grow, and access; and I think it shows better the intent of the data :)

roobre: I'd also add a small comment here explaining what this is for.

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, ed := range entitySynthesisDefinitions {
+		ed.tagRules = ed.Tags.getTagRules()
+		s.addConditions(ed.EntityRule)
+
+		for _, er := range ed.Rules {
+			er.EntityType = ed.EntityType
+			er.tagRules = append(er.Tags.getTagRules(), ed.EntityRule.tagRules...)
+			s.addConditions(er)
+		}
+	}
```
roobre, line 32: I think the short names (`ed`, `er`) make this code a bit difficult to follow. In this case I'd prefer to have longer but more descriptive names, like `definition` and `rule`.

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+	for k, v := range t {
+		if v == nil {
+			tr = append(tr, tagRule{name: k, entityTagName: k})
+			continue
+		}
+		if newName, ok := v["entityTagName"]; ok {
+			newNameS, _ := newName.(string)
+			tr = append(tr, tagRule{name: k, entityTagName: newNameS})
+		}
+	}
```
roobre, line 99: Similarly as above, I think short names here hurt readability, specially `k` and `v` (since it's not super obvious what the keys and values of this map actually model).

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (t Tags) getTagRules() []tagRule {
+	var tr []tagRule
+	for k, v := range t {
+		if v == nil {
+			tr = append(tr, tagRule{name: k, entityTagName: k})
+			continue
+		}
+		if newName, ok := v["entityTagName"]; ok {
```
roobre, line 95: What is `entityTagName`? Is this magic string a part of the entity synthesys protocol, or a convention for something else?

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+	return strings.HasPrefix(attribute, c.Prefix)
+}
+
+// Tags key value attributes
+type Tags map[string]map[string]interface{}
+
+type tagRule struct {
+	name          string
+	entityTagName string
+}
```
roobre, line 86: If I understood this correctly, `name` is the name of the attribute that contains the to-be-created tag value, and `entityTagName` is the name of this same tag after applying the renaming logic in the entity synthesis rule. Is this correct? If so, it might be better to clarify this either with a comment, or in the names of these fields:
```suggestion
type tagRule struct {
	sourceAttribute string
	newTagName      string
}
```

roobre (changes requested):
I've also found that there are many statements "cuddled" with each other (without empty lines in between). It's usually good practice to add those.

The https://github.com/mvdan/gofumpt and https://github.com/bombsimon/wsl can fix those for you :)

On a different note, it would seem that the `Synthesizer` object is well contained and with a nicely isolated responsibility. Would it be possible to move it to it's own package? i think the `integration` one is quite big already :(

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+
+// EntityRule contains rules to synthesis an entity
+type EntityRule struct {
+	EntityType string      `mapstructure:"type"`
+	Identifier string      `mapstructure:"identifier"`
+	Name       string      `mapstructure:"name"`
+	Conditions []Condition `mapstructure:"conditions"`
+	Tags       Tags        `mapstructure:"tags"`
+	tagRules   []tagRule
+}
```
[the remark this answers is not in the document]
gsanchezgavier: Tags can contain rules like renaming, this struct contains all tags of the rule preprocessed at initialization so it is easier to apply those rules on adding tags. 
I will add the comment too

internal/integration/entity_synthesis.go
gsanchezgavier: you're right, this is probably something that i left for a first implementation where i needed to look for the rules, I will change it for the list approach

internal/integration/entity_synthesis.go
gsanchezgavier: yes `entityTagName` is defined in the entity synthesis repo protocol, i forgot to add a comment there.

internal/integration/entity_synthesis.go
gsanchezgavier: I followed the entity synthesis protocol naming here, I think it will make easier to read if you are already familiar with the protocol , which is indeed needed for understand this package. I fear that doing more renaming will cause more confusion at the end.

internal/integration/entity_synthesis.go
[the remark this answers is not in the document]
gsanchezgavier: I made some changes on the suggested docs mainly for the condition.

internal/integration/entity_synthesis.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (rule EntityRule) isValid() bool {
+	return rule.Identifier != "" && rule.Name != "" && rule.EntityType != ""
+}
+
+// Condition contains parameters used to match entities from metrics
```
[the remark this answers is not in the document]
gsanchezgavier: changed for `Condition decides whether a metric is suitable to be included into an entity based on the metric attributes and metric name.`

roobre (approved):
LGTM, thanks for taking care of the comments!

PR #208, opened 2021-06-29
feat(infra_emitter): add common attributes

gsanchezgavier:
This PR fixes https://github.com/newrelic/newrelic-coreint/issues/109

It also removes the following attributes from the metrics since they are not needed on the sdkV4 protocol:
- `nrMetricType`
- `promMetricType`
- `integrationVersion` this was replaced by `instrumentation.version` populated in the Agent from the Integration Metadata
- `integrationName`  this was replaced by `instrumentation.name` populated in the Agent from the Integration Metadata


ivancorrales (commented):
_

roobre (approved):
LGTM!

The only thing I'd add is an integration test to ensure that the produced metrics no longer have `targetName` and friends as non-common attributes.

PR #210, opened 2021-07-02
feat: add integration_metadata configuration

gsanchezgavier:
Closes https://github.com/newrelic/newrelic-coreint/issues/114

### DIFF
--- a/internal/cmd/scraper/scraper.go
+++ b/internal/cmd/scraper/scraper.go
@@ -209,7 +209,7 @@ func RunOnceWithEmitters(cfg *Config, emitters []integration.Emitter) error {
 		)
 	}
 
-	// fetch duration is hardcoded to 1 since the target is scraped only once
+	// Fetch duration is hardcoded to 1 since the target is scraped only once
 	integration.ExecuteOnce(
 		retrievers,
 		integration.NewFetcher(scrapeDuration, cfg.ScrapeTimeout, cfg.WorkerThreads, cfg.BearerTokenFile, cfg.CaFile, cfg.InsecureSkipVerify, queueLength),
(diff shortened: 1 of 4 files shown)


internal/integration/infra_sdk_emitter.go
```diff
@@ -29,12 +29,33 @@ var removedAttributes = map[string]struct{}{
...
+
+// Metadata contains the name and version of the exporter that is being scraped.
+// The Infra-Agent use the metadata to populate instrumentation.name and instrumentation.value
+type Metadata struct {
+	Name    string `mapstructure:"name"`
+	Version string `mapstructure:"version"`
+}
+
+func (im *Metadata) isValid() bool {
+	return im.Name != "" && im.Version != ""
```
paologallinaharbur, line 44: should we log something in case one of the two is populated?

internal/cmd/scraper/scraper.go
```diff
@@ -313,7 +313,7 @@ func Run(cfg *Config) error {
 			emitters = append(emitters, emitter)
 		case "infra-sdk":
 			s := synthesis.NewSynthesizer(cfg.EntityDefinitions)
-			emitter := integration.NewInfraSdkEmitter(s)
+			emitter := integration.NewInfraSdkEmitter(cfg.IntegrationMetadata, s)
```
paologallinaharbur, line 316: codeclimate is complaining about something here I do not understand

paologallinaharbur (approved):

paologallinaharbur:
Two minor things, everything else looks good!

internal/integration/infra_sdk_emitter.go
roobre: We could perform this check when the metadata is being replaced and throw an error.

internal/cmd/scraper/scraper.go
gsanchezgavier: agree thanks for seeing this. changed

paologallinaharbur (approved):
It looks good!

Issue #150, opened 2021-03-17
Completed pods are getting scraped


anandkumarpatel:
Completed pods are getting scraped

## Description
When nri-prometheus looks for objects to scrape it does not take into account pod status. If all processes in a pod are stopped and will never be started again they should not be scraped. 

## Expected Behavior
Pods with all processes stopped should not be scraped. 

## Steps to Reproduce
Create a pod with `RestartPolicy` = `Never` and have the process exit.

Notice logs 
```
 prometheus-scraper time="2021-03-17T18:54:20Z" level=warning msg="fetching Prometheus metrics: http://100.96.95.84:15090/stats/prometheus (completed-container)" component=Fetcher error="Get \"http://100.96.95.84:15090/stats/
 prometheus-scraper time="2021-03-17T18:54:20Z" level=warning msg="error while scraping target" component=Fetcher error="Get \"http://100.96.95.84:15090/stats/prometheus\": dial tcp 100.96.95.84:15090: connect: no route to host"
```

## Your Environment
KinD, Minikube, EKS

PR: https://github.com/newrelic/nri-prometheus/pull/149

davidgit:
This issue was closed because it has been inactive for a long time despite it can be reopened at any point if you think it still relevant.

Thank you for your contribution!

PR #212, opened 2021-07-16
do not scrape pods that are completed

anandkumarpatel:
Pods that are completed and will never be restarted should not be scrapped because there is no process alive to handle the call. 

This PR add a check for pod completion to the `isObjectScrapable` function. 

I made it a switch statement in case there is a need to use object-specific keys to determine scrapability in the future. 

[Pod Phases doc](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase)
[Restart Policy doc](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy)

Fix for: https://github.com/newrelic/nri-prometheus/issues/150

Reopening: https://github.com/newrelic/nri-prometheus/pull/149

### DIFF
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -195,6 +195,13 @@ func (k *kubernetesTargetRetriever) listServices() error {
 }
 
 func isObjectScrapable(o metav1.Object, label string) bool {
+	switch obj := o.(type) {
+	case *corev1.Pod:
+		if isPodCompleted(obj) {
+			return false
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


paologallinaharbur:
Hello, first of all thanks for proposing the change. 😄 

The reason of the delay in the answer is that the repo and whole implementation is currently under discussion @gsanchezgavier  and we did not want to ask you to work on a PR that possibly was not going to be merged because of that

Therefore, we decided that for now we are not going to accept any new feature since likely we are going to move towards a different architecture.

PR #216, opened 2021-08-05
upgrade dependencies, build environment, and docker base image

roobre:
This PR bumps base images, build environment, and other dependencies


go.mod
```diff
@@ -3,22 +3,17 @@ module github.com/newrelic/nri-prometheus
 go 1.16
```
roobre, line 3: This needs a bump as well, but it's not yet in homebrew's repo.

paologallinaharbur (approved):

paologallinaharbur:
@roobre That dependency has been updated already in https://github.com/newrelic/nri-prometheus/pull/220/files

Is still useful this PR even if the main objective changed?

PR #235, opened 2021-10-14
allow metric filtering by suffix

smcavallo:
allow metric filtering by suffix
Ex.
```
transformations:
  - description: "General processing rules"
    ignore_metrics:
      - suffixes:
        - "_bucket"
```

The use case is the ability to ignore histograms which generate a large number of (unused) metrics.
```
# HELP argocd_redis_request_duration Redis requests duration.
# TYPE argocd_redis_request_duration histogram
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="0.01"} 2.791635e+06
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="0.05"} 3.347797e+06
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="0.1"} 3.582787e+06
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="0.25"} 3.609236e+06
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="0.5"} 3.609581e+06
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="1"} 3.609594e+06
argocd_redis_request_duration_bucket{hostname="argocd-application-controller-1",initiator="argocd-application-controller",le="+Inf"} 3.609602e+06
argocd_redis_request_duration_sum{hostname="argocd-application-controller-1",initiator="argocd-application-controller"} 41456.31533212083
argocd_redis_request_duration_count{hostname="argocd-application-controller-1",initiator="argocd-application-controller"} 3.609602e+06
```

In the above case we want to ignore `argocd_redis_request_duration_bucket` but keep metrics for `argocd_redis_request_duration_sum` and `argocd_redis_request_duration_count`

Signed-off-by: smcavallo <smcavallo@hotmail.com>

### DIFF
--- a/internal/integration/fetcher.go
+++ b/internal/integration/fetcher.go
@@ -26,7 +26,7 @@ import (
 
 // Fetcher provides fetching functionality to a set of Prometheus endpoints
 type Fetcher interface {
-	// Fetcher fetches data from a set of Prometheus /metrics endpoints. It ignores failed endpoints.
+	// Fetch fetches data from a set of Prometheus /metrics endpoints. It ignores failed endpoints.
 	// It returns each data entry from a channel, assuming this function may run in background.
 	Fetch(t []endpoints.Target) <-chan TargetMetrics
 }
--- a/internal/integration/rules.go
+++ b/internal/integration/rules.go
@@ -27,12 +27,13 @@ type RenameRule struct {
 	Attributes   map[string]interface{} `mapstructure:"attributes"`
 }
 
-// IgnoreRule skips for processing metrics that match any of the Prefixes.
+// IgnoreRule skips for processing metrics that match any of the Prefixes or Suffixes.
 // Metrics that match any of the Except are never skipped.
-// If Prefixes is empty and Except is not, then all metrics that do not
+// If Prefixes and Suffixes are empty and Except is not, then all metrics that do not
 // match Except will be skipped.
 type IgnoreRule struct {
 	Prefixes []string `mapstructure:"prefixes"`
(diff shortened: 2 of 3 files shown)


smcavallo:
@gsanchezgavier - codeclimate is complaining _Method ignoreRules.shouldIgnore has 5 return statements (exceeds 4 allowed)_
IMO changing the number of return statements in that method would reduce readability of that code and possibly performance.  Can this be excluded, and if so, where best to make this change - .codeclimate.yml or console?

smcavallo:
@gsanchezgavier - thanks for taking a look!  I'm not sure if I'm following what you're saying about the implementation - hoping you can help.   Is there an existing way to ignore every metric where the name ends with `_bucket` ?  We don't want to add ignore rules for prefixes with all our metric names which may have histograms - there are too many and it would be too many rules to maintain.

gsanchezgavier:
Sure ! a few weeks ago i debugged an issue where a prefix filter was being add for the same reason you mention here, filtering the `metric_name_bucket` and that was not working. During that debugging i checked that the filters where applied to the metric name which is `metric_name` for the example i'm giving. 
So filter the _bucket metrics of an histogram metric is not supported by this filters in the way is implemented now.

gsanchezgavier:
I will close this PR , feel free to open a FR issue for adding filters to the different time series of a histogram metric type.

PR #240, opened 2021-11-11
Add support for `prometheus.io/scheme`

### LINKED ISSUE #234: Add support for `prometheus.io/scheme` annotation/label to support https endpoints
## Is your feature request related to a problem? Please describe.
Auto discovered endpoints/services only supports `http` scheme. So `https` endpoints needs to be configured in the fixed targets despite they could be auto discovered and the correct scheme selected by using the `prometheus.io/scheme` [annotation](https://github.com/prometheus-community/helm-charts/blob/33c68f2c818e33723a0c3297bbe2488429f4b447/charts/prometheus/values.yaml#L1491).

## Feature Description
if `prometheus.io/scheme` is present in the endpoint/service annotation honor it in order to scrape that target.

## Describe Alternatives
right now a manual target can be configured.

## Additional context
Add any other context here.

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
[Nice to Have, Really Want, Must Have, Blocker]

vihangm:
- Adds scheme scraping with a deafult set to `http`.
- Cleans up the endpoints/service/pod target logic to go though the same
  pathways.
- Avoids unnecessary sprintf based URL creation and parsing.
- Test changes:
  - Bad annotations now cause a fallback to default scraping instead of
    returning nothing.
  - Not using sprintf for URL creation properly escapes paths. So
    `prometheus.io/path` should be a path and shouldn't include query
    params. `?` in paths get URL escaped now.

Fixes #234

Signed-off-by: Vihang Mehta <vihang@pixielabs.ai>

### DIFF
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -123,11 +123,7 @@ func nodeTargets(n *corev1.Node) ([]Target, error) {
 		return nil, err
 	}
 
-	lbls := labels.Set{}
-	for lk, lv := range n.Labels {
-		lbls["label."+lk] = lv
-	}
-
+	lbls := getK8sLabels(n)
 	for ty, a := range addrMap {
 		ln := "node_address_" + string(ty)
 		lbls[ln] = a[0]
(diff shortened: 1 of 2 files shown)


vihangm:
TODO: Add test cases for `prometheus.io/scheme` label and annotation overrides.

roobre:
HI @vihangm , thanks a lot for giving this a swing! We'll do our best to review this ASAP, please ping us whenever you think it's ready :)

vihangm:
@roobre Added some tests that I think should provide sufficient extra coverage. Good for a review!

vihangm:
Note on the failing checks: 
- The containerd CVE needs to be fixed by a dep https://github.com/newrelic/infrastructure-agent/pull/660
- The codeclimate duplication issue is pre-existing, it triggered because I moved some funcs for readability
- The coverage dropped because the LOC dropped due to refactoring but the uncovered lines stayed the same and the covered lines were refactored into reusable components.

gsanchezgavier (approved):
Thanks LGTM ! I did a quick [local test ](https://onenr.io/0Y8wpAooNRO) where discovery looks good . Just left a small suggestion.

PR #246, opened 2022-01-14
fix(infraSDKEmitter): Add tags as entity metadata

gsanchezgavier:
The agent distinguishes two kind of tags:
* tags that come in the telemetry and that aren’t changeable by the user - it calls it `metadata`. And this are the same that are defined in the [entity synthesis definitions](https://github.com/newrelic/entity-definitions/blob/main/definitions/infra-powerdns_authoritative/definition.yml#L12) as `tags`.

* tags that are defined by the user, either as [custom attributes](https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings#custom-attributes) in the agent or as [labels](https://docs.newrelic.com/docs/infrastructure/host-integrations/infrastructure-integrations-sdk/specifications/host-integrations-standard-configuration-format/#labels) in the integration config. Those are handled by the Agent. 

We were adding the attributes as entity tags instead of metadata what caused the attributes names to be prefixed with `tags.` and making the backend synthesis to fail.

### DIFF
--- a/internal/synthesis/synthesis.go
+++ b/internal/synthesis/synthesis.go
@@ -64,7 +64,7 @@ type EntityRule struct {
 	Identifier string      `mapstructure:"identifier"` // Name of the attribute that will be used to uniquely identify the synthesized entity.
 	Name       string      `mapstructure:"name"`       // Name of the attribute that will be used to name the synthesized entity.
 	Conditions []Condition `mapstructure:"conditions"` // List of rules used to determining if a metric belongs to this entity.
-	Tags       Tags        `mapstructure:"tags"`       // List of attributes that will be added to the entity as tags.
+	Tags       Tags        `mapstructure:"tags"`       // List of attributes that will be added to the entity as metadata.
 	tagRules   []tagRule   // List of all tags better structured to facilitate synthesis process.
 }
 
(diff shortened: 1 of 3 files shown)


internal/synthesis/synthesis.go
```diff
@@ -64,7 +64,7 @@ type EntityRule struct {
 	Identifier string      `mapstructure:"identifier"` // Name of the attribute that will be used to uniquely identify the synthesized entity.
 	Name       string      `mapstructure:"name"`       // Name of the attribute that will be used to name the synthesized entity.
 	Conditions []Condition `mapstructure:"conditions"` // List of rules used to determining if a metric belongs to this entity.
-	Tags       Tags        `mapstructure:"tags"`       // List of attributes that will be added to the entity as tags.
+	Tags       Tags        `mapstructure:"tags"`       // List of attributes that will be added to the entity as metadata.
```
roobre, line 67: Should we keep this old name? Changing might be a bit painful but will make everything more clear for our future selves.

roobre:
A random general thought: We are replacing tags support by metadata support. It is then implied that we won't need tags in the foreseeable future, is this right?

gsanchezgavier:
I think that Tags are only added by the agent when custome attributes or labels are configured and this 2 aggregations are done in the agent after receiving the integration payload so i would remove them from the sdk if there is no use case for that   , avoiding this kinds of bugs

internal/synthesis/synthesis.go
gsanchezgavier: This structure follows the naming conventions of the entity definition files so I would keep it like this to avoid more confusion since we will be renaming what entity definitions describe in the entity rules. On line 144 there is the actual conversion of naming between entity definitions and infra sdk.
Any way this hole code has a short expectation of live since the hole entity synthesis process will be removed from pomi

gsanchezgavier:
this will be closed in favor of  #250

PR #253, opened 2022-02-04
Add prometheus accept header

### LINKED ISSUE #252: bug with URL encoded params returned

## Description

since release `2.10.0` using URL parameters doesn't work anymore. For example with `2.9.0` having the annotation `prometheus.io/path: /metrics?format=prometheus` worked, since `2.10.0` it queries it as `/metrics%3Fformat=prometheus`.

This was once solved in #46 but returned.

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )
`GET /metrics?format=prometheus`

## Steps to Reproduce
[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )
Set the annotation `prometheus.io/path: /metrics?format=prometheus`

## Your Environment
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )
Running on AKS

## Additional context
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )
#46

gsanchezgavier:
Adds the `Accept` header with the same content as the Prometheus Server.

Removes the "application/json" content-type form the header as well since only prometheus text metrics can be parsed by this Getter 

fixes #252

### DIFF
--- a/internal/pkg/prometheus/prometheus_test.go
+++ b/internal/pkg/prometheus/prometheus_test.go
@@ -5,6 +5,7 @@ package prometheus_test
 import (
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


paologallinaharbur (approved):

internal/pkg/prometheus/prometheus_test.go
```diff
@@ -5,45 +5,32 @@ package prometheus_test
...
-# HELP http_requests_total Total number of HTTP requests made.
-# TYPE http_requests_total counter
-http_requests_total{code="200",handler="prometheus",method="get"} 2
-`
-
 func TestGet(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		_, _ = w.Write([]byte(result))
+		accept := r.Header.Get("Accept")
+		if !strings.Contains(accept, "application/openmetrics-text;") {
```
roobre, line 19: I'd get rid of the trailing `;`, just in case it changes in the future
```suggestion
		if !strings.Contains(accept, "application/openmetrics-text") {
```

roobre (approved):

PR #256, opened 2022-02-19
Feature: Add support for annotation ignore metrics

danifbento:
At this moment all ignored metrics should be set on the nri-prometheus configuration. This PR adds the possibility of ignoring a set of metrics using a annotation per pod/service.

Signed-off-by: Daniela Bento <danibento@overdestiny.com>

### DIFF
--- a/internal/cmd/scraper/scraper.go
+++ b/internal/cmd/scraper/scraper.go
@@ -173,6 +173,7 @@ func RunWithEmitters(cfg *Config, emitters []integration.Emitter) error {
 		retrievers,
 		integration.NewFetcher(scrapeDuration, cfg.ScrapeTimeout, cfg.WorkerThreads, cfg.BearerTokenFile, cfg.CaFile, cfg.InsecureSkipVerify, queueLength),
 		integration.RuleProcessor(processingRules, queueLength),
+		integration.AnnotationRulesProcessor,
 		emitters)
 
 	r := http.NewServeMux()
@@ -214,6 +215,7 @@ func RunOnceWithEmitters(cfg *Config, emitters []integration.Emitter) error {
 		retrievers,
 		integration.NewFetcher(scrapeDuration, cfg.ScrapeTimeout, cfg.WorkerThreads, cfg.BearerTokenFile, cfg.CaFile, cfg.InsecureSkipVerify, queueLength),
 		integration.RuleProcessor(cfg.ProcessingRules, queueLength),
+		integration.AnnotationRulesProcessor,
 		emitters)
 
 	return nil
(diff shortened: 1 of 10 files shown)


roobre:
Hey @danifbento, thanks a lot for your work here!

We've briefly discussed this internally, and we'd like know more about the use case you had in mind regarding this. What was the behavior you were trying to get, and why weren't you able to achieve it with the current configuration facilities?

Our current stance regarding `nri-prometheus` is for it to be as simple as possible, and we want to carefully consider whether we should implement features that are not availabe on OSS tooling (like prometheus server itself), as doing so would hurt customer's ability to switch from `nri-prometheus` to prometheus server.

danifbento:
Hello @roobre 

Before all, thanks for replying. 

The reasoning behind this PR was prevent sending not needed metrics into New Relic, avoiding sending to much data into the platform (as we don't need those metrics) and also it avoid so many traffic for each report. Regarding this topic we can always try to setup a transformation on nri-prometheus configuration to ignore all those metrics we don't want to be pushed into the system.

But, the rational was also another, for example, ArgoCD Controller expose tons of metrics in it `/metrics` endpoint. The majority we don't need them (ArgoCD Controller is only one example, but we can have more, of course). But for instance, to drop those metrics for ArgoCD Controller we will drop all metrics that match for that rule for all the services which are being scrapped by nri-prometheus (for what I understood of the implementation of the scrapper). Which we don't want to do, example, we can filter out all GoLang metrics, but for some services we want those metrics. So, a general config implies more challenge on how to scrap and what to scarp. 

This PR borns of the ideia if we can say to a service if it should be scrapped or not, why cannot we configure each service to say which metrics should be reported or not? Because each service knows exactly what metrics it is exposing, so in the point of view of maintainability it will be easier to put this effort on the service itself. If there is no annotation, all metrics should be scrapped as normal.

For instance, it was for this cascading of precedences that I put the AnnotationRulesProcessor to be run after the normal processor, because we should go from general rules into specific rules of each service.

In case of ArgoCD Controller we also reduce the payload in a very extreme way, which is less traffic and less unneeded metrics on New Relic.

I hope to make myself well explained.

alvarocabanas (commented):
Hi @danifbento, we are considering your proposal and will discuss it internally with other aspects for the nri-prometheus integration.

danifbento:
Hello @alvarocabanas, thanks for considering it.

davidgit:
We appreciate you taking the time to bring this to our attention. After reviewing this issue, we are unable to commit on a timeline to resolution because we are already working on a replacement for nri-prometheus that would not be compatible with the solution you proposed

If the situation changes, we will revisit this decision.

Thanks for your contribution!

PR #263, opened 2022-03-18
Bump SDK version, remove entity synthesis and add ignoreEntity flag.

alvarocabanas:
- Bump SDK version
- Remove entity synthesis
- Add ignoreEntity flag.


internal/cmd/scraper/scraper.go
```diff
@@ -51,13 +50,12 @@ type Config struct {
 	EmitterProxy                      string `mapstructure:"emitter_proxy"`
 	// Parsed version of `EmitterProxy`
 	EmitterProxyURL                              *url.URL
-	EmitterCAFile                                string                 `mapstructure:"emitter_ca_file"`
-	EmitterInsecureSkipVerify                    bool                   `mapstructure:"emitter_insecure_skip_verify" default:"false"`
-	TelemetryEmitterDeltaExpirationAge           time.Duration          `mapstructure:"telemetry_emitter_delta_expiration_age"`
-	TelemetryEmitterDeltaExpirationCheckInterval time.Duration          `mapstructure:"telemetry_emitter_delta_expiration_check_interval"`
-	WorkerThreads                                int                    `mapstructure:"worker_threads"`
-	EntityDefinitions                            []synthesis.Definition `mapstructure:"entity_definitions"`
```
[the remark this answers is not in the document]
gsanchezgavier: we might want to print a warn with deprecation notice WDYT?

internal/integration/infra_sdk_emitter.go
```diff
@@ -80,6 +77,11 @@ func (e *InfraSdkEmitter) Emit(metrics []Metric) error {
 	if err != nil {
 		return err
 	}
+	// We want the agent to not send metrics attached to any entity in order to make the entity synthesis to take place
+	// completely in the backend. Since V4 SDK still needs an entity (Dataset) to attach the metrics to, we are using
+	// the default hostEntity to attach all the metrics to it but setting this flag, IgnoreEntity: true that
+	// will cause the agent to send them unattached to any entity
+	i.HostEntity.SetIgnoreEntity(true)
```
gsanchezgavier, line 84: would make sense to add a test to check metrics are added to a dataset without entity metadata?

internal/cmd/scraper/scraper.go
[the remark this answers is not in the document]
alvarocabanas: Yes it's a good Idea, in case it's easy to detect, I would emit the warn, at least I will add the test.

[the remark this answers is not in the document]
alvarocabanas: Done, with the warning

internal/pkg/endpoints/endpoints.go
```diff
@@ -99,14 +104,25 @@ func urlToTarget(URL string, TLSConfig TLSConfig) (Target, error) {
...
 		TLSConfig: TLSConfig,
 		URL:       *u,
 	}, nil
 }
+
+// ReplaceLocalhost replaces the occurrence of a localhost address with
+// the given hostname.
+func replaceLocalhost(source, with string) string {
+	return localhostReplaceRE.ReplaceAllString(source, with)
+}
```
alvarocabanas, line 128: This is a copy of the function existing on the SDK. I Thought about asking CAOS to make it public so we can call it from here without redeclaring it. I see points in favor and against. First, right now this function leaves on the integration package of the SDK, optimally, if exporting it, I would ask CAOS to place it in a utils package or something. The point that I see against using an exported one is in the case that we needed the two functions to diverge. What do you thing @gsanchezgavier ?

internal/cmd/scraper/scraper.go
```diff
@@ -116,12 +116,12 @@ func RunWithEmitters(cfg *Config, emitters []integration.Emitter) error {
 		return fmt.Errorf("you need to configure at least one valid emitter")
 	}
```
alvarocabanas, line 118: If we added ctx context to this RunWithEmitters and allowed the server creation to be done creting the server instead of calling the helper function [listenAndServe](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/server.go;drc=b5e06545b3c3b3b9c2ed8204e5ba3fd25a85db0f;l=3220) I would be able to put a go routine listening to the context cancel and forcing the server to shutdown, so we can create a test like the one there is for RunOnceWithEmitters, but I understand that, since we are thinking on refactoring all the integration, it's better to touch the less as posible and just test this manually by running the binary, do you agree @gsanchezgavier ?

internal/integration/infra_sdk_emitter.go
alvarocabanas: I don't fully understand the case, the infra_sdk_emitter_test it's already testing a case in  Test_Emitter_EmitsEntity where no entity metadata is added but it's in the attributes. Do we need an extra one?

internal/pkg/endpoints/endpoints.go
alvarocabanas: Oks, I added tests covering the different cases for replacement, we will stick to this local copy.

internal/cmd/scraper/scraper.go
```diff
@@ -116,12 +116,12 @@ func RunWithEmitters(cfg *Config, emitters []integration.Emitter) error {
...
 
-	selfRetriever, err := endpoints.SelfRetriever()
+	selfRetriever, err := endpoints.SelfRetriever(cfg.HostID)
```
gsanchezgavier, line 119: I would add the replacement mechanism only in the infra_sdk_emitter but not in the endpoint packages that affects the standalone function of POMI. I don't think this is a feature we want to have in k8s since there should be not localhost targetsNames AFAIK

internal/integration/infra_sdk_emitter.go
gsanchezgavier: oh probably i missed when I check , not needed to duplicate if the test exist

internal/cmd/scraper/scraper.go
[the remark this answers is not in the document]
gsanchezgavier: uff this is messy, i don't know why review comments are getting duplicated. I will resolve this and leave the other open

cmd/nri-prometheus/config.go
```diff
@@ -54,6 +56,10 @@ func loadConfig() (*scraper.Config, error) {
 		return nil, errors.Wrap(err, "could not read configuration")
 	}
 
+	if cfg.Get("entity_definitions") != nil {
+		logrus.Warn("entity_definitions are deprecated and won't be processed since v2.14.0")
```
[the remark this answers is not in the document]
alvarocabanas: I changed it to debug.

internal/cmd/scraper/scraper.go
alvarocabanas: Moved to the emitter

internal/integration/infra_sdk_emitter_test.go
```diff
@@ -380,33 +306,165 @@ redis_foo_test{hostname="localhost",env="dev",uniquelabel="test"} 3
...
+	testMetadata := Metadata{
+		Name:    "nri-foo",
+		Version: "test",
+	}
+
+	for _, c := range cases {
+		c := c
+
+		t.Run(c.testName, func(t *testing.T) {
+
```
gsanchezgavier, line 441: not actually needed but if possible t.parallel should be used.

gsanchezgavier (commented):
LGTM I just left some nits in the tests

internal/integration/infra_sdk_emitter_test.go
```diff
@@ -380,33 +306,165 @@ redis_foo_test{hostname="localhost",env="dev",uniquelabel="test"} 3
...
+			input: TargetMetrics{
+				Metrics: []Metric{
+					{
+						name:       "a-metric",
+						value:      float64(3),
+						metricType: "count",
+						attributes: labels.Set{
+							"targetName":        "128.0.0.1:8080",
+							"scrapedTargetName": "128.0.0.1:8080",
+							"scrapedTargetURL":  "https://128.0.0.1:8080",
```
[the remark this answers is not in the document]
alvarocabanas: I want to be sure it's not changed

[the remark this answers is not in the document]
alvarocabanas: If on the future it's changed because of a change in the code, I want the test to fail.

internal/integration/infra_sdk_emitter_test.go
alvarocabanas: t.Parallel() is intended to be used for slow tests only; average tests are so fast that any gain from parallel execution would be negligible. I really don't know why we are using it everywhere if go test -race is used.

alvarocabanas: It actually tends to mess up table driven tests adding extra complexity for no reason.

internal/integration/infra_sdk_emitter_test.go
```diff
@@ -380,33 +306,165 @@ redis_foo_test{hostname="localhost",env="dev",uniquelabel="test"} 3
...
+	cases := []struct {
+		testName     string
+		input        TargetMetrics
+		expectedName string
+		expectedURL  string
+		hostID       string
+	}{
+		{
+			testName: "When hostID provided but host not localhost",
+			input: TargetMetrics{
```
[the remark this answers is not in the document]
alvarocabanas: Ok sure.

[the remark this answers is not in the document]
alvarocabanas: done

internal/integration/infra_sdk_emitter_test.go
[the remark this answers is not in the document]
gsanchezgavier: It might be good to leave a comment with the reason why , if it's something that do not need to be changed

PR #270, opened 2022-03-29
Release, lint, test charts

paologallinaharbur:


.github/ct.yaml
```diff
@@ -0,0 +1,4 @@
+# Chart linter defaults to `master` branch so we need to specify this as the default branch
+# or `cl` will fail with a not-so-helpful error that says:
+# "Error linting charts: Error identifying charts to process: Error running process: exit status 128"
```
roobre, line 3: ❤️ the comment

.github/workflows/push_pr.yml
```diff
@@ -14,6 +14,37 @@ env:
...
+            echo "::set-output name=changed::true"
+          fi
+      - name: Run helm unit tests
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
+          helm plugin install https://github.com/quintush/helm-unittest
+          for chart in $(ct --config .github/ct.yaml list-changed); do
+            if [ -d "$chart/tests/" ]; then
+              helm unittest -3 $chart
+            fi
```
roobre, line 45: Can we add a message if the chart has to tests? Might help with debugging later.

```suggestion
            if [ -d "$chart/tests/" ]; then
              helm unittest -3 $chart
            else
              echo "No unit tests found for $chart"
            fi
```

.github/workflows/release-chart.yaml
```diff
@@ -0,0 +1,40 @@
...
+  validate-gh-pages-index:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+      - name: Download remote index file and check equality
+        run: |
+          curl -vsSL https://newrelic.github.io/nri-prometheus/index.yaml > index.yaml.remote
```
roobre, line 19: Can we use Github magic vars/expressions to get `nri-prometheus` procedurally? I think it would be useful for forks etc.

.github/ct.yaml
paologallinaharbur: credits to @kang-makes

roobre (approved):

kang-makes (approved):
Although I left a NIT, it looks good to me.

I would make an offline simply question to @roobre if he is comfortable with having `load-test-environment` chart directly in `charts` instead of `charts/internal` like we had in [`nri-kubernetes`](https://github.com/newrelic/nri-kubernetes/tree/main/charts/internal/e2e-resources)

charts/load-test-environment/templates/deployment.yaml
```diff
@@ -4,14 +4,16 @@
...
 {{- $index = add1 $index -}}
 {{- $latency := default "0" .latency -}}
-{{- $latencyVariation := default "0" .latencyVariation   -}}
-{{- $indexString := toString $index  -}}
+{{- $latencyVariation := default "0" .latencyVariation -}}
+{{- $indexString := toString $index -}}
+
+{{- $uniqueDeployName := printf "%s-lat%s-latvar%s-index%s" .name $latency $latencyVariation $indexString }}
+
+---
```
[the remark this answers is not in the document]
paologallinaharbur: I fixed it removing the newline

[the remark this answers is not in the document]
kang-makes: It was because you put an end line chomper at the end so the next iteration of the `range` does not have a newline: https://github.com/newrelic/nri-prometheus/pull/270/files/91f213dcca0d565c92d0b46e4fdbe25682d5f9b7#diff-812ad1bdf11f045fde10edcd67619f2ba9e99b886e87734f42a5d69672d0bb99R69

PR #273, opened 2022-04-08
feat(cicd): test install and upgrade of charts

paologallinaharbur:
Linked to https://github.com/newrelic/newrelic-coreint/issues/314

Fixing as well security pipeline and bumping image

### DIFF
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -19,15 +19,16 @@ jobs:
     name: Helm chart Lint
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    strategy:
+      matrix:
+        kubernetes-version: [ "v1.16.15", "v1.22.0" ]
     steps:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
       - uses: helm/chart-testing-action@v2.2.1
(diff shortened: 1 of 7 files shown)


.github/workflows/push_pr.yml
```diff
@@ -46,6 +47,31 @@ jobs:
               echo "No unit tests found for $chart"
             fi
           done
+      - name: Setup Minikube
+        uses: manusa/actions-setup-minikube@v2.4.3
+        if: steps.list-changed.outputs.changed == 'true'
+        with:
+          minikube version: v1.20.0
```
gsanchezgavier, line 54: nit: i think there could be some newer versions

gsanchezgavier (approved):
LGTM

.github/workflows/push_pr.yml
paologallinaharbur: you are right, I don't even know why we hard code that, I'll bump it or remove it. Thanks for noticing

paologallinaharbur: we cannot remove it since it is required, but we can bump it

kang-makes (approved):
LGTM

PR #276, opened 2022-04-13
Replace helpers with common library

gsanchezgavier:
This PR removes all custom helpers the chart had that were replaced by the common library.

No breaking changes should be introduced in the values.

List of changes that might see after upgrade:

- `app.kubernetes.io/managed-by`, `app.kubernetes.io/version`, `helm.sh/chart` pod labels being removed (To be checked)


charts/nri-prometheus/tests/labels_test.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - equal:
+          path: spec.template.metadata.labels.[app.kubernetes.io/instance]
+          value: release
+        template: templates/deployment.yaml
+      - equal:
+          path: spec.template.metadata.labels.[app.kubernetes.io/name]
+          value: nri-prometheus
+        template: templates/deployment.yaml
+
+## TODO check possible breaking, podLabels is not adding this labels which were added by common lables.
```
gsanchezgavier, line 48: @kang-makes found this difference when implementing the podLabels , was it contemplated to remove these labels with the common library?

charts/nri-prometheus/Chart.yaml
```diff
@@ -1,19 +1,36 @@
...
-home: https://github.com/newrelic/nri-prometheus
-icon: https://newrelic.com/assets/newrelic/source/NewRelic-logo-square.svg
+description: A Helm chart to deploy the New Relic Prometheus OpenMetrics integration
+
+home: https://docs.newrelic.com/docs/infrastructure/prometheus-integrations/install-configure-openmetrics/configure-prometheus-openmetrics-integrations/
+icon: https://newrelic.com/themes/custom/curio/assets/mediakit/new_relic_logo_vertical.svg
+sources:
+  - https://github.com/newrelic/nri-prometheus
+
+version: 2.0.0
```
gsanchezgavier, line 11: just realized that there this should be just a minor bump, I will change it when addressing more changes request from reviews

charts/nri-prometheus/templates/clusterrolebinding.yaml
```diff
@@ -2,15 +2,15 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "nri-prometheus.fullname" . }}
+  name: {{ template "newrelic.common.naming.fullname" . }}
```
sigilioso, line 5: I'm not sure if we have any preference using `template` or `include`, but if we did, this would be a go chance to go in one direction or another. WDYT?

sigilioso (commented):
LGTM, I just left a comment which we may discuss

However, I think having an additional review from someone whose eyes are more hardened to helm charts would be a good idea 👀

charts/nri-prometheus/templates/clusterrolebinding.yaml
gsanchezgavier: ups totally miss this , I think it should be [include](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-include-function)  Also the common library examples are with include. I will fix this on all the chart, thanks

charts/nri-prometheus/templates/deployment.yaml
```diff
@@ -1,26 +1,25 @@
...
   namespace: {{ .Release.Namespace }}
   labels:
-{{ include "nri-prometheus.labels" . | indent 4 }}
+    {{- include "newrelic.common.labels" . | nindent 4 }}
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app.kubernetes.io/name: {{ include "nri-prometheus.name" . }}
+      app.kubernetes.io/name: {{ include "newrelic.common.naming.name" . }}
```
gsanchezgavier, line 12: just found that there is a special template function for this :) I will change it

charts/nri-prometheus/templates/clusterrolebinding.yaml
gsanchezgavier: done added in new commit

sigilioso: 👍 thanks!

sigilioso:
Just a remainder: I would [update common.library to 1.0.1](https://github.com/newrelic/helm-charts/pull/784)

gsanchezgavier:
> Just a remainder: I would https://github.com/newrelic/helm-charts/pull/784
@sigilioso  i got a PR in the oven ready to come out, i will do it there

sigilioso (approved):

charts/nri-prometheus/tests/labels_test.yaml
```diff
@@ -0,0 +1,58 @@
+suite: test object names
```
kang-makes, line 1: As I already said, this is redundant to the test we do in the common library. I would remove them an use a `ci/test-values.yaml` to "simply" test that the chart installs.

charts/nri-prometheus/tests/configmap_test.yaml
```diff
@@ -19,3 +19,67 @@ tests:
             scrape_services: true
             transformations: []
             verbose: false
+
+  - it: creates the config map with lowDataMode.
```
kang-makes, line 23: Now I understand when you quote Mateusz: "I only have to read the tests"

I was not able to understand the logic (and I know that it was me who merge this from the community)

Good test 👍

.github/workflows/push_pr.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         with:
           fetch-depth: 0
       - uses: helm/chart-testing-action@v2.2.1
+      - name: Add newrelic repository
+        run: helm repo add newrelic https://helm-charts.newrelic.com
```
kang-makes, line 31: NIT/something to discuss: 

I would leave this as it is because there are more than two other repos that have the repository added like this BUT there is another (more cleaner if you ask me) way to add this that @sigilioso [has in its PR](https://github.com/newrelic/k8s-metadata-injection/blob/bb45f8a0c1d49fd47fa96bb95cd14b8603ce327f/.github/ct.yaml#L6-L7) but un the `ct.yaml` file.

We should decide which way we should standardize this. (In fact this is a job I have to do)

charts/nri-prometheus/Chart.yaml
```diff
@@ -8,7 +8,7 @@ icon: https://newrelic.com/themes/custom/curio/assets/mediakit/new_relic_logo_ve
 sources:
   - https://github.com/newrelic/nri-prometheus
 
-version: 2.0.0
+version: 1.15.0
```
kang-makes, line 11: I put this above in the review but I think this is a breaking change because of Helm 2 deprecation... But I am asking in the Slack channel to talk about this as a team.

kang-makes (commented):
We talked you were going to create another PR with more things from the common library so I see things like `hostNetwork` or `dnsConfig` unimplemented yet.

Take a minute to address the comments I left and give me a glance at the second PR to know if I am missing something in this "first season" or if it a cliff hanger 🤣

kang-makes (approved):
Re-reading my comments, all of them are NITs.

Approved. Thank you!

charts/nri-prometheus/tests/metadata_name_test.yaml
```diff
@@ -0,0 +1,40 @@
+suite: test object names
```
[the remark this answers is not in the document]
gsanchezgavier: yes, this was at the very begging of the work were i didn't have a clear idea of what we should test here (I'm don't still have 🤣 ) . I implemented tests before changing the templates to have an idea of what could brake, but now looked like trivial tests and for sure this is already tested in the library. 

I will get rid of them.

charts/nri-prometheus/tests/labels_test.yaml
gsanchezgavier: done, removed

charts/nri-prometheus/templates/clusterrolebinding.yaml
```diff
@@ -11,6 +11,6 @@ roleRef:
   name: {{ template "common.naming.fullname" . }}
 subjects:
 - kind: ServiceAccount
-  name: {{ template "nri-prometheus.serviceAccountName" . }}
+  name: {{ include "common.serviceAccount.name" . }}
```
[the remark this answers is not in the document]
gsanchezgavier: oh i totally missed that , thanks. I just added

charts/nri-prometheus/tests/labels_test.yaml
gsanchezgavier: yes, ct test are already in place, i removed these tests.

.github/workflows/push_pr.yml
gsanchezgavier: I changed to follow the same approach for constancy, on the other hand it is making the add repo step a bit hidden for the unit test step ( i think that is also needed for that step if i'm not wrong)

PR #288, opened 2022-05-03
endpoints/kubernetes: parse query args from prometheus.io/path

roobre:
As prometheus server abuses label concatenation using regexes, the `prometheus.io/path` annotation can be abused to include query paramenters such as `prometheus.io/path: /metrics?format=foo`, and this will be handled fine by it.

While this behavior is likely accidental, it has been set as standard and some edge cases rely upon it.

This PR allows `nri-prometheus` to support this use case, by pre-parsing the `prometheus.io/path` annotation into `path` and `query` before building the `url.URL` object.

### DIFF
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -249,6 +249,20 @@ func getPort(o metav1.Object) string {
 	return ""
 }
 
+// parsePath parses a raw URL query, such as `/metrics?format=foo` and returns separately the path and the query.
+// This is needed because the `prometheus.io/path` annotation is often abused, and query arguments are included in it.
+// As HTTP clients will escape out any `?` and `=` present in the path, we need to split it first to form a correct URL.
+func parsePath(pathQuery string) (string, string) {
+	parts := strings.Split(pathQuery, "?")
+	if len(parts) != 2 {
+		// Annotation contents either do not contain query arguments, or contain a malformed query.
+		// In either case we refuse to parse it as a query and handle it to url.URL to sanitize it.
+		return pathQuery, ""
+	}
+
+	return parts[0], parts[1]
+}
+
 func endpointsTarget(e *corev1.Endpoints, u url.URL) Target {
 	lbls := getK8sLabels(e)
 	// Name and Namespace of services and endpoints collides
@@ -270,7 +284,7 @@ func endpointsTarget(e *corev1.Endpoints, u url.URL) Target {
 func endpointsTargets(e *corev1.Endpoints, s *corev1.Service) []Target {
 	// we need to pass the service since the annotations are not inherited
 	port := getPort(s)
-	path := getPath(s)
+	path, query := parsePath(getPath(s))
 	scheme := getScheme(s)
 
 	var targets []Target
@@ -288,9 +302,10 @@ func endpointsTargets(e *corev1.Endpoints, s *corev1.Service) []Target {
 					continue
 				}
 				u := url.URL{
-					Scheme: scheme,
-					Host:   net.JoinHostPort(eSubAddr.IP, subPortStr),
-					Path:   path,
+					Scheme:   scheme,
+					Host:     net.JoinHostPort(eSubAddr.IP, subPortStr),
+					Path:     path,
+					RawQuery: query,
 				}
 				targets = append(targets, endpointsTarget(e, u))
 			}
@@ -319,7 +334,7 @@ func serviceTarget(s *corev1.Service, u url.URL) Target {
 func serviceTargets(s *corev1.Service) []Target {
 	port := getPort(s)
 	scheme := getScheme(s)
-	path := getPath(s)
+	path, query := parsePath(getPath(s))
 
 	if port != "" {
 		u := url.URL{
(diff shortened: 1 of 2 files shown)


internal/pkg/endpoints/kubernetes.go
```diff
@@ -270,7 +284,7 @@ func endpointsTarget(e *corev1.Endpoints, u url.URL) Target {
 func endpointsTargets(e *corev1.Endpoints, s *corev1.Service) []Target {
 	// we need to pass the service since the annotations are not inherited
 	port := getPort(s)
-	path := getPath(s)
+	path, query := parsePath(getPath(s))
```
paologallinaharbur, line 287: getPath name is still valid? should we update somehow the signature?

internal/pkg/endpoints/kubernetes.go
```diff
@@ -249,6 +249,20 @@ func getPort(o metav1.Object) string {
 	return ""
 }
 
+// parsePath parses a raw URL query, such as `/metrics?format=foo` and returns separately the path and the query.
+// This is needed because the `prometheus.io/path` annotation is often abused, and query arguments are included in it.
+// As HTTP clients will escape out any `?` and `=` present in the path, we need to split it first to form a correct URL.
+func parsePath(pathQuery string) (string, string) {
+	parts := strings.Split(pathQuery, "?")
```
paologallinaharbur, line 256: could we do something like https://go.dev/play/p/m8JLsUiYEHO?

paologallinaharbur: the example is not perfect since I just adapted an existing one

paologallinaharbur (commented):
LGTM but we could maybe improve a bit the splitting.

internal/pkg/endpoints/kubernetes_test.go
```diff
@@ -2184,6 +2184,117 @@ func TestPodTargetsPathLabel(t *testing.T) {
...
+	t.Parallel()
+
+	assert.ElementsMatch(
+		t,
+		podTargets(&corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "my-pod",
+				Namespace: "test-ns",
+				Annotations: map[string]string{
+					"prometheus.io/path": "/metrics?format=prometheus",
```
paologallinaharbur, line 2197: would be helpful to have 1-2 cases more, or even just this, but with a more complex example?

internal/pkg/endpoints/kubernetes_test.go
```diff
@@ -2184,6 +2184,117 @@ func TestPodTargetsPathLabel(t *testing.T) {
...
+						"deploymentName": "",
+						"nodeName":       "node-a",
+					},
+				},
+				URL: url.URL{
+					Scheme: "http",
+					Host:   "10.0.0.1:80",
+					// This is fine, url.URL will escape this to /metrics%3Fformat%3Dprometheus%3Fsomething%0A
+					// upon making the request.
+					Path: "/metrics?format=prometheus?something",
```
paologallinaharbur, line 2291: This sounds like a featurbug

internal/pkg/endpoints/kubernetes.go
roobre: Might me bending corners a little bit but as the label/annotation called `prometheus.io/path` I'm not super unhappy about the name.

I'd love to get rid of the `get` prefix but I did not want to refactor too much either :(

paologallinaharbur: fine for me

internal/pkg/endpoints/kubernetes_test.go
roobre: It was, such as putting query parameters in `prometheus.io/path` 😭 
I've reworked it to be more strict now.

internal/pkg/endpoints/kubernetes.go
roobre: Done, I think it's cleaner now :)

internal/pkg/endpoints/kubernetes_test.go
roobre: Now that we use `url.Parse`, do you think it's necessary? I think it would be testing `url.Parse` rather than our code.

paologallinaharbur: now it is not!

paologallinaharbur (approved):
Thanks for the changes!

internal/pkg/endpoints/kubernetes.go
```diff
@@ -249,6 +249,20 @@ func getPort(o metav1.Object) string {
...
+// abused, and query arguments are included in it.
+// If the contents of prometheus.io/path cannot be parsed as a URL, it returns an error, which will skip the target.
+// Note that url.Parse will not return an error for an empty rawUrl.
+func parsePath(pathQuery string) (string, string, error) {
+	u, err := url.Parse(pathQuery)
+	if err != nil {
+		return "", "", fmt.Errorf("cannot parse url from %q: %w", pathQuery, err)
+	}
+
+	return u.Path, u.RawQuery, nil
```
paologallinaharbur, line 263: you could also return u and use it as base object and add to it Host and Scheme..

But having the parsePath as builder of the url would be super weird 😓

roobre: I thought about this, refactoring this further and having a URL builder given an object. However this was very hard given how entagled the code is now :(

The main reason I did not return the URL was to avoid having to copy it for different iterations of the for loop, dereferencing the pointer for copying would have look bad :(

PR #289, opened 2022-05-03
Adding autogenerated README

kang-makes:


charts/nri-prometheus/values.yaml
```diff
@@ -220,41 +194,51 @@ config:
...
-# This proxy won't be used when scraping metrics from the targets.
-# By default it's empty, meaning that no proxy will be used.
-proxy:
-
-# Whether the integration should run in verbose mode or not.
-# Default: false
+  # fedramp.enabled -- (bool) Enables FedRAMP. Can be configured also with `global.fedramp.enabled`
+  # @default -- false
+  enabled:
+# -- (bool) Reduces number of metrics sent in order to reduce costs. Can be configured also with `global.lowDataMode`
```
paologallinaharbur, line 234: would make sense to ask the user to check the readme for further info? Or copy them directly here?

paologallinaharbur: Thinking about it, maybe it is enough to have them there, it is called readme for a reason 😅

charts/nri-prometheus/values.yaml
```diff
@@ -220,41 +194,51 @@ config:
...
-
-# Pod scheduling priority
-# Ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
-# priorityClassName: high-priority
-
+# -- Annotations to be added to all pods created by the integration.
+podAnnotations: {}
+
+
+# These values are managed by the common-library
```
paologallinaharbur, line 201: also for this chart I do not see much value into creating a different section for this

paologallinaharbur (approved):

charts/nri-prometheus/values.yaml
kang-makes: I cannot duplicate all the documentation 😅

charts/nri-prometheus/values.yaml
kang-makes: I applied here the suggestions from newrelic/nri-kube-events#112

kang-makes: Solved in [#4c27761](https://github.com/newrelic/nri-prometheus/pull/289/commits/4c27761bd01799514b1cb1a35cf6d70f77565200) and [#cc2ab29](https://github.com/newrelic/nri-prometheus/pull/289/commits/cc2ab298fbfd49ff712d8351be912afbed4685dd)

PR #301, opened 2022-05-24
fix(query): query params not included if port specified

paologallinaharbur:
Query params were not included whenever a port was specified


roobre (approved):
Thanks for catching this!

roobre (approved):

Issue #211, opened 2021-07-07
Add nightly builds 


invidian:
Similar to https://github.com/newrelic/nri-kubernetes/pull/134, it would be good to add support for creating nighty Docker images, so they can be deployed to the canary environments.

PR #303, opened 2022-05-26
Add nightly builds

marcsanmi:
Closes https://github.com/newrelic/nri-prometheus/issues/211.

This PR adds a new GHA workflow, which will build and push Docker images daily.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,46 @@
+name: Nightly build
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  push:
+    branches:
+      - main
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GO_VERSION: '1.17'
+  INTEGRATION: "prometheus"
+  ORIGINAL_REPO_NAME: 'newrelic/nri-prometheus'
+  REPO_FULL_NAME: ${{ github.event.repository.full_name }}
+  TAG: nightly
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Upload artifact for docker build step
+        uses: actions/upload-artifact@v2
+        with:
+          retention-days: 1
+          name: nri-prometheus-${{ matrix.goos }}-${{ matrix.goarch }}
+          path: bin/nri-prometheus-${{ matrix.goos }}-${{ matrix.goarch }}
+  docker:
+    name: Build and push docker images
+    needs: [ build ]
+    runs-on: ubuntu-latest
```
roobre, line 46: This uses `docker buildx` while the real release workflow uses goreleaser: https://github.com/newrelic/nri-prometheus/blob/main/.github/workflows/release-integration.yml#L74=

I think both workflows should be as similar as possible to avoid having one working while the other does not.

marcsanmi: Sure, going to update it. Now that you mention it... Should we try to unify across all repos as much as possible the release workflow?

roobre: We should! But `nri-prometheus` is special, because we need both a multiarch docker image, but also multiple binaries per arch, as you can run it on-host.

Using goreleaser to also produce docker images was easier than making the usual docker pipeline also produce binaries. I don't think the result is ideal (complexity is substantially higher), but given that we'll likely get rid of this solution soon... I'm not sure it makes sense to put more time on refactoring the pipeline :)

roobre (approved):

PR #309, opened 2022-06-07
Fix misleading comment

### LINKED ISSUE #306: Remove comment "  # If lowDataMode is se to true the config.transformations option will be ignored"
I do not believe this comment is still valid: https://github.com/newrelic/nri-prometheus/blob/main/charts/nri-prometheus/values.yaml#L171

Looks like we do a merge in the helper functions now:
https://github.com/newrelic/nri-prometheus/blob/main/charts/nri-prometheus/templates/configmap.yaml
https://github.com/newrelic/nri-prometheus/blob/main/charts/nri-prometheus/templates/_helpers.tpl

kang-makes:
Fixes #306


sigilioso (approved):
👍 thanks for taking care of it

Issue #315, opened 2022-06-30
errors fetching Prometheus metric - text format parsing error 


joncolby:
We are seeing the following errors for some scraped endpoints after switching to the nri-prometheus helm chart (2.1.5) ,  previously using nri-prometheus from the pre-V2 registry `https://helm-charts.newrelic.com` 

```
time="2022-06-30T07:53:32Z" level=warning msg="fetching Prometheus metrics: http://10.82.219.6:9001/prometheus (<REDACTED>)" component=Fetcher error="text format parsing error in line 348: invalid escape sequence '\\\"'"
time="2022-06-30T07:53:32Z" level=warning msg="error while scraping target" component=Fetcher error="text format parsing error in line 348: invalid escape sequence '\\\"'"
time="2022-06-30T07:53:32Z" level=debug msg="fetching URL: {http   10.82.136.207:9001 /prometheus  false   }" component=Fetcher target=<REDACTED>
time="2022-06-30T07:53:32Z" level=warning msg="fetching Prometheus metrics: http://10.82.136.207:9001/prometheus (<REDACTED>)" component=Fetcher error="text format parsing error in line 157: invalid escape sequence '\\\"'"
```

## Description

Examining the prometheus output manually,  there are no such escape sequences to be seen.  I cannot share the prometheus output due to confidentiality issues.

## Expected Behavior

no such errors during nri-prometheus parsing of targets

paologallinaharbur:
Hello, likely it is related to https://github.com/newrelic/nri-prometheus/issues/313

To doublecheck could you check the prometheus output and tell us if you notice differences querying with:

 - No Accept header
 - Header ‘Accept: application/openmetrics-text;version=0.0.1,text/plain;version=0.0.4;q=0.5,/;q=0.1’
-  Header ’Accept: text/plain;version=0.0.4;q=1,/;q=0.1’

joncolby:
Hi @paologallinaharbur -   I am happy to help debug this.

How do you mean I should query the prometheus targets with these different Headers?   Using curl?   I probably won't notice a different in visual output.    Because this problem happens even when there are no fields containing escape sequences `\`  in the prometheus metrics output.

Or is there a way to configure the nri-prometheus agent to use different headers when scraping targets?

paologallinaharbur:
Yes I meant with curl! 
No, but we are going to address this issue soon.

[...]

Issue #313, opened 2022-06-23
NRI-Prometheus not compatible with Open Metrics 1.0.0 / Spring-Boot


fieldju:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )
NRI Prometheus will not scrape Springboot services.
NRI Prometheus only knows how to deserialize open-metrics `v0.0.1` but current impls are `1.0.0`

This header [here](https://github.com/newrelic/nri-prometheus/blob/main/internal/pkg/prometheus/prometheus.go#L40): 

`application/openmetrics-text;version=0.0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1`

Is out of step with [Prometheus official](https://github.com/prometheus/prometheus/blob/v2.36.2/scrape/scrape.go#L755):

`application/openmetrics-text;version=1.0.0,application/openmetrics-text;version=0.0.1;q=0.75,text/plain;version=0.0.4;q=0.5,*/*;q=0.1`

Springboot seems to return `application/openmetrics-text;version=1.0.0;charset=utf-8` regardless of the version specified in the accept header.

if you drop `application/openmetrics-text` from the accept springboot returns `text/plain;version=0.0.4;charset=utf-8` which nri-prometheus does know how to process.

But by default NRI-Prometheus throws `text format parsing error in line 2: invalid escape sequence '\"'` when processing `application/openmetrics-text;version=1.0.0`

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )
I expect that NRI-Prometheus can consume `application/openmetrics-text;version=1.0.0` or configures the `Accept` header in a way that the Springboot Prometheus endpoint can handle.

## Steps to Reproduce
[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )

:point_down: here is a test that reproduces the issue

```patch
Index: internal/pkg/prometheus/prometheus_test.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/internal/pkg/prometheus/prometheus_test.go b/internal/pkg/prometheus/prometheus_test.go
[...]

PR #317, opened 2022-07-07
fix(acceptHeader): rollback to text/plain

paologallinaharbur:
The purpose of the PR is to remove from the accept header the format we actually do not support such as `openmetrics` and `*/*`. At the same time 

Related to https://github.com/newrelic/nri-prometheus/issues/252. In order to support a specific use-case we added as acceptHeader the very same value the prometheus server had. However, the prometheus server has two different parsers one for `plain/text` and the other one for `application/openmetrics-text`.

We currently leverage `expfmt.FmtText` to formatdata. `github.com/prometheus/common/expfmt` also supports `application/openmetrics-text` but merely up to 0.0.1

Fix https://github.com/newrelic/nri-prometheus/issues/313 
Fix https://github.com/newrelic/nri-prometheus/issues/315

---

For example configuring:
```
    scrape_accept_header: "application/openmetrics-text;version=0.0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1"
```

having the fake openmetric exporter (by @gsanchezgavier 💟 ) running it fails:
```
time="2022-07-07T12:53:38Z" level=warning msg="fetching Prometheus metrics: http://172.17.0.9:8000/metrics (test-resources-openmetrics-858864575d-9fssr)" component=Fetcher error="text format parsing error in line 70: unknown metric type \"info\""
time="2022-07-07T12:53:38Z" level=warning msg="error while scraping target" component=Fetcher error="text format parsing error in line 70: unknown metric type \"info\""

```

with the default one it works

### DIFF
--- a/cmd/nri-prometheus/config.go
+++ b/cmd/nri-prometheus/config.go
@@ -94,6 +94,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("require_scrape_enabled_label_for_nodes", true)
 	viper.SetDefault("scrape_timeout", 5*time.Second)
 	viper.SetDefault("scrape_duration", "30s")
+	viper.SetDefault("scrape_accept_header", "text/plain;version=0.0.4")
 	viper.SetDefault("emitter_harvest_period", fmt.Sprint(integration.BoundedHarvesterDefaultHarvestPeriod))
 	viper.SetDefault("min_emitter_harvest_period", fmt.Sprint(integration.BoundedHarvesterDefaultMinReportInterval))
 	viper.SetDefault("max_stored_metrics", fmt.Sprint(integration.BoundedHarvesterDefaultMetricsCap))
@@ -105,7 +106,6 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("scrape_endpoints", false)
 	viper.SetDefault("percentiles", []float64{50.0, 95.0, 99.0})
 	viper.SetDefault("worker_threads", 4)
-	viper.SetDefault("accept_header", "text/plain;version=0.0.4")
 }
 
 // bindViperEnv automatically binds the variables in given configuration struct to environment variables.
--- a/cmd/nri-prometheus/config.go
+++ b/cmd/nri-prometheus/config.go
@@ -105,6 +105,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("scrape_endpoints", false)
 	viper.SetDefault("percentiles", []float64{50.0, 95.0, 99.0})
 	viper.SetDefault("worker_threads", 4)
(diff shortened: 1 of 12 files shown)


gsanchezgavier (approved):
thanks! LGTM

sigilioso (approved):
👍 awesome!

cmd/nri-prometheus/config.go
```diff
@@ -94,6 +94,7 @@ func setViperDefaults(viper *viper.Viper) {
 	viper.SetDefault("require_scrape_enabled_label_for_nodes", true)
 	viper.SetDefault("scrape_timeout", 5*time.Second)
 	viper.SetDefault("scrape_duration", "30s")
+	viper.SetDefault("scrape_accept_header", "text/plain;version=0.0.4")
```
roobre, line 97: Should this be an exposed config value? I think we should send the correct header for the things that the code supports. Users modifying this will almost surely lead to broken behavior.

roobre (changes requested):
I think the header we were using was _almost_ correct:

We were using:
`application/openmetrics-text;version=0.0.1,text/plain;version=0.0.4;q=0.5,*/*;q=0.1`

And I believe our parser should be able to handle `application/openmetrics-text;version=0.0.1,text/plain;version=0.0.4` just right.

The problem is when we get sent `application/openmetrics-text;version=1.0.0`, which our current header allows as per the latest part: `*/*;q=0.1`, meaning "Ok I will accept anything with less preference than the things I said before"

roobre:
Left [a comment](https://github.com/newrelic/nri-prometheus/issues/313#issuecomment-1178863228) on the issue, I think this is a bug in the server implementation and we should think very carefully before we make this change in our (what I believe to be) correct code to match a buggy implementation.

paologallinaharbur:
For sure '*/*' is wrong since we do not support everything. With the currently implementation we are not able to scrape a fake exporter with the python client

cmd/nri-prometheus/config.go
paologallinaharbur: Talking offline we decided to change the header and support the value of the one in the prometheus server previous to the open-metric change 

https://github.com/prometheus/prometheus/commit/9c03e11c2cf2ad6c638567471faa5c0f6c11ba3d

roobre (approved):
It seems that I was mistaken by believing that OpenMetrics 0.0.1 was compatible with the parser we used, and OpenMetrics 0.0.1 was not. Turns out this is not true: OpenMetrics 0.0.1 was not compatible to begin with and should have never been in our `Accept` header.

https://github.com/newrelic/nri-prometheus/issues/313#issuecomment-1178926951

LGTM now, I'd only suggest to use the exact `Accept` header that Prometheus server used before the OpenMetrics parser was added.

PR #319, opened 2022-07-11
feat(chart): pointing to image version 2.16.2

paologallinaharbur:
Pointing to new release https://github.com/newrelic/nri-prometheus/releases/tag/v2.16.2

### DIFF
--- a/charts/nri-prometheus/README.md
+++ b/charts/nri-prometheus/README.md
@@ -1,6 +1,6 @@
 # nri-prometheus
 
-![Version: 2.1.5](https://img.shields.io/badge/Version-2.1.5-informational?style=flat-square) ![AppVersion: 2.16.1](https://img.shields.io/badge/AppVersion-2.16.1-informational?style=flat-square)
+![Version: 2.1.7](https://img.shields.io/badge/Version-2.1.7-informational?style=flat-square) ![AppVersion: 2.16.2](https://img.shields.io/badge/AppVersion-2.16.2-informational?style=flat-square)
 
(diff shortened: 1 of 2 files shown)


charts/nri-prometheus/README.md
```diff
@@ -1,6 +1,6 @@
 # nri-prometheus
 
-![Version: 2.1.5](https://img.shields.io/badge/Version-2.1.5-informational?style=flat-square) ![AppVersion: 2.16.1](https://img.shields.io/badge/AppVersion-2.16.1-informational?style=flat-square)
+![Version: 2.1.7](https://img.shields.io/badge/Version-2.1.7-informational?style=flat-square) ![AppVersion: 2.16.2](https://img.shields.io/badge/AppVersion-2.16.2-informational?style=flat-square)
```
gsanchezgavier, line 3: I think we talked about this last week, and it could be wiped from the template.

gsanchezgavier (approved):

charts/nri-prometheus/README.md
kang-makes: If we had this automatic I would not bother but as we don't have it is a toil that I don't like to have.

We can remove it from the template here: https://github.com/newrelic/nri-prometheus/blob/65e51542d7/charts/nri-prometheus/README.md.gotmpl#L4

kang-makes (approved):

charts/nri-prometheus/README.md
paologallinaharbur: yes we would need to fix them across all charts 😢

PR #327, opened 2022-07-25
feat(useBearer): expose the config for static targets

paologallinaharbur:
Attaching the token to all request would be a security issue. 

Therefore the decision was taken to attach it merely for static targets whenever the user was explicitly asking for it.


gsanchezgavier (approved):
LGTM , would make sense to add some test?

paologallinaharbur:
Sure, I added them!

PR #333, opened 2022-07-28
chore: use infra-checkers static analysis settings and bump go to  1.18

sigilioso:


.github/workflows/push_pr.yml
```diff
@@ -78,13 +78,18 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-go@v2
```
paologallinaharbur, line 81: There is GO_VERSION env variable, can you also update it at line 60 and elsewhere?

By the way I asked in many PRs, so sorry if you read it already, why do we need to setup go now?

paologallinaharbur (approved):
Just a nit and small doubts regarding the way go is setup

.github/workflows/push_pr.yml
sigilioso: Sure, thanks!

sigilioso: Including this comment to keep in handy: Needed to use golangci-lint-action v3. Review [this](https://github.com/golangci/golangci-lint-action/pull/403) for details.

paologallinaharbur (approved):

PR #340, opened 2022-09-08
Update README.md

marcsanmi:
Add missing release name in `helm upgrade --install` command.


sigilioso (commented):
Shouldn't we update it in [README.md.gotmpl](https://github.com/newrelic/nri-prometheus/blob/main/charts/nri-prometheus/README.md.gotmpl#L15) and use `helm-docs`?

sigilioso (approved):
Thank you 👍

Issue #344, opened 2022-09-20
Ignore all metrics except set of metrics


alvarocabanas:
As a user, I want to only collect a defined set of metric names and exclude everything else. 

```
transformations:
  - description: allow
     ignore_metrics:
      - except:
        - apiserver_audit_requests_rejected_total
```

This behavior, even if it's contemplated in the tests, it's not working properly.

paologallinaharbur:
Trying to reproducing it locally.
having:
```
  transformations:
    - description: allow
      ignore_metrics:
        - except:
            - go
```

seems to be working as expected
<img width="961" alt="Screenshot 2022-09-23 at 16 11 21" src="https://user-images.githubusercontent.com/43335750/191980185-86b6d204-d09c-4091-8ef5-d202f2258acd.png">

Moreover, there is a test covering that use case and the code seems without issues.
```go
		exceptRulesLen += len(rule.Except)
		for _, prefix := range rule.Except {
			if strings.HasPrefix(name, prefix) {
				return false
			}
		}
[...]
		// Prefixes
		prefixesLen += len(rule.Prefixes)
		for _, prefix := range rule.Prefixes {
			if strings.HasPrefix(name, prefix) {
				return true
			}
		}
```

I am trying to verify if we reproduced it locally at some point

paologallinaharbur:
The issue is related to lowDatamode and the way multiple rules are managed.

```
	if prefixesLen > 0 || metricTypesLen > 0 {
		return false
	}
```
This is considered at the end of all rules, therefore it is enough that one had a prefix to make useless a rule with only excludes


Having:
```
transformations:
- description: allow
  ignore_metrics:
  - except:
    - apiserver_audit_requests_rejected_total
- description: Low data mode defaults
  ignore_metrics:
  - prefixes:
    - kube_
    - container_
    - machine_
    - cadvisor_
```

Then becomes similar to:
```
- description: allow
  ignore_metrics:
  - except:
    - apiserver_audit_requests_rejected_total
  - prefixes:
    - kube_
    - container_
    - machine_
    - cadvisor_
```

PR #346, opened 2022-09-26
NEWRELIC-3719 fix(filters): multiple rules bug

paologallinaharbur:
Fix https://github.com/newrelic/nri-prometheus/issues/344

The previous implementation was counting prefixes and metricTypes from all rules when applying the excludes.

With the old implementation the testcase TestIgnoreRules was starting to fail adding a prefix without any effect to a different rule.

The risk of solving this bug is breaking the config of users relying on a bug to collect metrics

### DIFF
--- a/internal/integration/rules.go
+++ b/internal/integration/rules.go
@@ -209,17 +209,19 @@ func addAttributes(targetMetrics *TargetMetrics, rules []AddAttributesRule) {
 type ignoreRules []IgnoreRule
 
 func (rules ignoreRules) shouldIgnore(name string, metricType metricType) bool {
-	var prefixesLen, metricTypesLen, exceptRulesLen int
+	if rules.isMetricExcludedFromFilters(name) {
+		return false
+	}
+
 	for _, rule := range rules {
-		exceptRulesLen += len(rule.Except)
-		for _, prefix := range rule.Except {
-			if strings.HasPrefix(name, prefix) {
-				return false
-			}
+		// if metricTypesLen or prefixesLen are not defined and exceptRulesLen
+		// is not empty then all not excepted metric should be dropped
+		totalDroppingRules := len(rule.MetricTypes) + len(rule.Prefixes)
+		if totalDroppingRules == 0 && len(rule.Except) != 0 {
+			return true
 		}
 
 		// MetricTypes
-		metricTypesLen += len(rule.MetricTypes)
 		for _, rMetricType := range rule.MetricTypes {
 			if strings.EqualFold(rMetricType, string(metricType)) {
 				return true
@@ -227,7 +229,6 @@ func (rules ignoreRules) shouldIgnore(name string, metricType metricType) bool {
 		}
 
 		// Prefixes
-		prefixesLen += len(rule.Prefixes)
 		for _, prefix := range rule.Prefixes {
 			if strings.HasPrefix(name, prefix) {
 				return true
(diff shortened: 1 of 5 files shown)


internal/integration/rules.go
```diff
@@ -209,38 +209,47 @@ func addAttributes(targetMetrics *TargetMetrics, rules []AddAttributesRule) {
...
-		exceptRulesLen += len(rule.Except)
-		for _, prefix := range rule.Except {
-			if strings.HasPrefix(name, prefix) {
-				return false
-			}
+		// if metricTypesLen or prefixesLen are not defined and exceptRulesLen
+		// is not empty then all not excepted metric should be dropped
+		totalDroppingRules := len(rule.MetricTypes) + len(rule.Prefixes)
+		if totalDroppingRules == 0 && len(rule.Except) != 0 {
+			return true
```
gsanchezgavier, line 221: is there any except rule it will ignore the metric without checking the prefix? is this ok?

internal/integration/rules_test.go
```diff
@@ -428,6 +428,17 @@ func TestAddAttributesRules(t *testing.T) {
 	}
 }
 
+func TestEmptyIgnoreRules(t *testing.T) {
```
gsanchezgavier, line 431: could you add the test case that reproduce the bug? It will make easier to understand the exact case

paologallinaharbur: I did it already adding `Prefixes: []string{"not_matching"},`. I can create one more specific if needed. Also in the issue there are some info

internal/integration/rules.go
gsanchezgavier: oh i get this now, at this point all `except` rules has executed, so if the metric is not excepted and there are not other rules it should be ignored.

internal/integration/rules.go
```diff
@@ -209,38 +209,47 @@ func addAttributes(targetMetrics *TargetMetrics, rules []AddAttributesRule) {
...
 		}
 	}
 
-	if prefixesLen > 0 || metricTypesLen > 0 {
-		return false
+	return false
+}
+
+// When matching an except rule we do not drop the metric, no matter if a rule is dropping it after
+func (rules ignoreRules) isMetricExcludedFromFilters(name string) bool {
```
gsanchezgavier, line 243: nit: I would change the terms exclude and filter for except and rules. 
`metricExcepted` or `metricExceptedByRules`

internal/integration/rules.go
```diff
@@ -209,38 +209,47 @@ func addAttributes(targetMetrics *TargetMetrics, rules []AddAttributesRule) {
 type ignoreRules []IgnoreRule
 
 func (rules ignoreRules) shouldIgnore(name string, metricType metricType) bool {
-	var prefixesLen, metricTypesLen, exceptRulesLen int
+	if rules.isMetricExcludedFromFilters(name) {
```
gsanchezgavier, line 212: perhaps add a comment here explaining why the except rules are executed before the rest of the rules.

gsanchezgavier (approved):
LGTM i left some nits. I would also add a description of the cases where this fix could change the behavior of a current transformation which we should add to the release notes

internal/integration/rules_test.go
paologallinaharbur: I added some further tests

internal/integration/rules.go
paologallinaharbur: true, in the code the field is .except

PR #361, opened 2022-10-21
Unprivileged Namespace Only scanner

jepries:
All, please consider adding a feature to scan only the namespace you are in.

Currently, the K8S io client interface is using "Cluster Scope" when scanning for pods and services.  In my environment, my containers are not allowed to have Cluster Scope so I've added a field to the config

`cluseter_namespace`

that if present, will use the namespace in the kubernetes client when scanning for pods and services.  Leaving it blank reverts to normal processing of using Cluster Scope when scanning for pods and services.

this feature only required 3 changes for me to get it to work:

k8s-target-retriever to be able to run hand tests against my namespace (This is a great script, thanks for it)
scraper.go to pull the cluster_namespace parameter from config and send it to
kubernetes.go to use it when scanning for pods and services

There may be more places this needs changing in the test classes?


roobre:
I think that for restricting which workloads should be monitored, a better option is to define a custom scrape annotation (i.e. _not_ `prometheus.io/scrape`), and stamp said workloads with it.

Is there any particular reason for which you would like to restrict by namespace?

roobre:
Hey John,

Thanks a lot for your detailed answer!
You are totally right, if you do not have permission to list things in the cluster scope, `nri-prometheus` will refuse to work.

This is an interesting use case, I will bring this up internally to see if we can make room for it. Currently we're focusing our efforts in https://github.com/newrelic/newrelic-prometheus-configurator/, which according to our plans should eventually replace `nri-prometheus`. I'll bring this up to the teams to see if that use case could be incorporated there.

jepries:
@roobre Thanks!

I'll also be working on another feature to reduce the amount of attributes that are sent with a metric.  I see we have the ability to add attributes (labels) but no feature to remove them.

I have to work on reducing the cardinality of the data sent and limiting the attributes (labels) will help me accomplish that..... 

This feature is available in prometheus "metric_relabel_configs", I'm going to add something similar.

I like the prometheus_configurator idea, i'll look into that!

gsanchezgavier:
Hi @jepries  i'm closing this PR since [prometheus configurator](https://github.com/newrelic/newrelic-prometheus-configurator) supports this and we are not longer adding new features to nri-prometheus.

PR #365, opened 2022-10-28
fix helm upgrade command

bpschmitt:
The `helm upgrade` command was invalid.  Fixing so that it works :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ You can install this chart using [`nri-bundle`](https://github.com/newrelic/helm
 
 ```shell
 helm repo add nri-prometheus https://newrelic.github.io/nri-prometheus
-helm upgrade --install nri-prometheus/nri-prometheus -f your-custom-values.yaml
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ You can install this chart using [`nri-bundle`](https://github.com/newrelic/helm
 
 ```shell
 helm repo add nri-prometheus https://newrelic.github.io/nri-prometheus
```
kang-makes, line 22: The repository that we are adding for Prometheus is `nri-prometheus` so the install should be `nri-prometheus/nri-prometheus`.

Am I missing something?

kang-makes (commented):
We cannot change the repository to `newrelic` because it will collide with other READMEs like the one used for the `nri-bundle`.

In fact, if the customer is installing the `nri-bundle` and have `newrelic` repo set from there, that command will install the deprecated chart.

Let me know if we are missing something.

bpschmitt:
@kang-makes this can be closed.  User error on my part :) I completely missed adding the `nri-prometheus` repo.

PR #381, opened 2023-02-24
chore: fix config example default values

sigilioso:
The default value for `scrape_timeout` in the config example comment is misleading because wasn't reflecting the [default value actually being used](https://github.com/newrelic/nri-prometheus/blob/chore/update-config-examples/cmd/nri-prometheus/config.go#L96).

### DIFF
--- a/configs/nri-prometheus-config.yml.sample
+++ b/configs/nri-prometheus-config.yml.sample
@@ -27,7 +27,7 @@ integrations:
       audit: false
 
       # The HTTP client timeout when fetching data from endpoints. Defaults to "5s" if it is not set.
-      # scrape_timeout: "30s"
+      # scrape_timeout: "5s"
 
       # Length in time to distribute the scraping from the endpoints. Default to "30s" if it is not set.
       scrape_duration: "5s"
--- a/configs/nri-prometheus-config.yml.sample
+++ b/configs/nri-prometheus-config.yml.sample
@@ -26,10 +26,10 @@ integrations:
       # It does not include verbose mode. This can lead to a high log volume, use with care.
       audit: false
 
-      # The HTTP client timeout when fetching data from endpoints. Defaults to 30s.
+      # The HTTP client timeout when fetching data from endpoints. Defaults to "5s" if it is not set.
       # scrape_timeout: "30s"
 
-      # Length in time to distribute the scraping from the endpoints.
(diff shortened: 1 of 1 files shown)


configs/nri-prometheus-config.yml.sample
```diff
@@ -26,11 +26,11 @@ integrations:
...
       audit: false
 
-      # The HTTP client timeout when fetching data from endpoints. Defaults to 30s.
-      # scrape_timeout: "30s"
+      # The HTTP client timeout when fetching data from endpoints. Defaults to 5.
+      # scrape_timeout: "5s"
 
       # Length in time to distribute the scraping from the endpoints.
-      scrape_duration: "5s"
+      scrape_duration: "30s"
```
paologallinaharbur, line 33: this is not commented out, therefore we are changing the default, is this intended?

sigilioso: You are totally right, I saw the `sample` suffix in the file name and I forgot the value can be used as the default (even if it is an example and not the "real" default). I'm restoring the default and including a comment to say that it defaults to that value if it is not set.

sigilioso: Updated here: [16b8496](https://github.com/newrelic/nri-prometheus/pull/381/commits/16b8496473292d102f5e35ec3823497a02b15d55)

configs/nri-prometheus-config.yml.sample
```diff
@@ -26,10 +26,10 @@ integrations:
       # It does not include verbose mode. This can lead to a high log volume, use with care.
       audit: false
 
-      # The HTTP client timeout when fetching data from endpoints. Defaults to 30s.
+      # The HTTP client timeout when fetching data from endpoints. Defaults to "5s" if it is not set.
       # scrape_timeout: "30s"
```
paologallinaharbur, line 30: since this is commented we can reflect the real value 5s

paologallinaharbur (approved):

configs/nri-prometheus-config.yml.sample
sigilioso: Changed here: [7c04876](https://github.com/newrelic/nri-prometheus/pull/381/commits/7c048766863e62e028c94944db44d7123ec9f015)

PR #392, opened 2023-03-18
Update CODEOWNERS

htroisi:


kang-makes (changes requested):
`nri-prometheus` has not been handed off and it is still managed (and deprecated) by Core Integrations.

PR #423, opened 2023-06-21
Upgrade Alpine Image Version

xqi-nr:
to 3.18.2
apline 3.18 has severe vulnerabilities


paologallinaharbur:
Alpine 3.18 is pointing currently to 3.18.2

Pinning the minor is a way to update automatically, however having the minor there could help release more frequently

paologallinaharbur (approved):

PR #425, opened 2023-06-22
chore(pomi): update appVersion

paologallinaharbur:
We were not updating the chart version due to https://github.com/renovatebot/renovate/issues/13968

We should understand if we want to move the version to the values.yaml or change the renovate config to support regexes

### DIFF
--- a/.github/workflows/load_test.yml
+++ b/.github/workflows/load_test.yml
@@ -39,6 +39,7 @@ jobs:
         with:
           minikube version: v1.30.1
           kubernetes version: v1.25.6
+          driver: docker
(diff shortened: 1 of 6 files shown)


xqi-nr (approved):

xqi-nr (approved):

xqi-nr (approved):

kang-makes (approved):
I saw tests passing so I expect tests to pass again as the last commit was only to remove the comment I left the last time I tried to do this.

I don't know if something has changed on the last version to allow this that I was not able to find, a lot of time has passed.

Nevermind, congratulations! Now we can upgrade our test pipelines :)

LGTM.

PR #458, opened 2023-11-13
Do not target services with port that does not exist

kirecek:
When `prometheus.io/port` annotation is set, no matter what, it just creates the target according the annotation value.

This potentially could make unnecessary calls to non-exposed port that end up with errors on every scrape tick.

```
... context deadline exceeded (Client.Timeout exceeded while awaiting headers) ...
```

This can be easily validated so this PR adds check for desired port and **throws a single error** if port does not exist.

_Note: It's user error, so correct solution is to get rid of annotation, however, annotation might be static in third party packages therefore unnecessary overrides are needed till it's get fixed. In each case, this cloud help to have cleaner logs when user error occurs._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Unreleased section should follow [Release Toolkit](https://github.com/newrelic/r
 
 ## Unreleased
 
+### Enhancement
+- Skip prometheus services with non-existing port
+
 ## v2.19.0 - 2023-11-02
 
 ### 🚀 Enhancements
--- a/internal/pkg/endpoints/kubernetes.go
+++ b/internal/pkg/endpoints/kubernetes.go
@@ -344,7 +344,16 @@ func serviceTargets(s *corev1.Service) []Target {
 		return nil
 	}
 
+	availablePorts := make(map[string]bool)
+	for _, port := range s.Spec.Ports {
+		availablePorts[fmt.Sprintf("%d", port.Port)] = true
+	}
+
 	if port != "" {
+		if !availablePorts[port] {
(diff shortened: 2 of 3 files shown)


sigilioso:
Hi @kirecek,  Thanks for your contribution!

One of the existing tests is failing, could you take a look at it?

You can execute tests locally using:

```bash
$ make test
```

Thanks!

kirecek:
oh, sure thing. Sorry, didn't notice it because I executed only newly added test 🤦🏻 

Should be fixed now.

```
make test
=== prometheus === [ test ]: running unit tests...
?   	github.com/newrelic/nri-prometheus/cmd/k8s-target-retriever	[no test files]
ok  	github.com/newrelic/nri-prometheus/cmd/nri-prometheus	(cached)
ok  	github.com/newrelic/nri-prometheus/internal/cmd/scraper	(cached)
ok  	github.com/newrelic/nri-prometheus/internal/integration	(cached)
?   	github.com/newrelic/nri-prometheus/internal/retry	[no test files]
?   	github.com/newrelic/nri-prometheus/load-test/mockexporter	[no test files]
ok  	github.com/newrelic/nri-prometheus/internal/pkg/endpoints	(cached)
ok  	github.com/newrelic/nri-prometheus/internal/pkg/labels	(cached)
ok  	github.com/newrelic/nri-prometheus/internal/pkg/prometheus	(cached)
```

Thank you :)

sigilioso:
Sorry the delay. There was an issue with the repository pipelines that is already fixed, could you rebase with main to bring the latest changes? Thanks!

kirecek:
It's totally fine.

Branch is rebased. Thank you!

sigilioso (approved):

PR #470, opened 2024-02-19
Feat/header xprometheus

paologallinaharbur:
Some exporters assume to receive this header in order to stop complex computations.
https://github.com/percona/mongodb_exporter/blob/6b436b67c9ee63745d20d9c059403185ad0b7aed/exporter/exporter.go#L273

Prometheus server is including it in all requests:
https://github.com/prometheus/prometheus/blob/ac10cd4d9966ae7a86c696cd7440ddce13f5c068/scrape/scrape.go#L712

### DIFF
--- a/internal/integration/fetcher.go
+++ b/internal/integration/fetcher.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


internal/integration/fetcher.go
```diff
@@ -253,7 +253,7 @@ func (pf *prometheusFetcher) fetch(t endpoints.Target) (prometheus.MetricFamilie
 		httpClient = pf.bearerClient
 	}
 
-	mfs, err := pf.getMetrics(httpClient, t.URL.String(), pf.acceptHeader)
+	mfs, err := pf.getMetrics(httpClient, t.URL.String(), pf.acceptHeader, fmt.Sprintf("%f", pf.fetchTimeout.Seconds()))
```
paologallinaharbur, line 256: I need to change this

internal/pkg/prometheus/prometheus.go
```diff
@@ -36,15 +36,23 @@ func ResetTargetSize() {
 	targetSize.Reset()
 }
 
+const (
+	// XPrometheusHeader included in all requests. It informs exporters about its timeout.
+	XPrometheusHeader = "X-Prometheus-Scrape-Timeout-Seconds"
```
sigilioso, line 41: Nit: should we use `XPrometheusScrapeTimeoutHeader` as constant name? It is quite a bit long but it is also more descriptive.

sigilioso (approved):
LGTM 👍 

Although I wonder if we should document somewhere this new behavior. Maybe we could extend the description in the [config example](https://github.com/newrelic/nri-prometheus/blob/main/configs/nri-prometheus-config.yml.sample#L29). WDYT?

sigilioso (approved):
Thanks! 👍

PR #491, opened 2024-07-15
feat(chart): updated chart

paologallinaharbur:
Changes in https://github.com/newrelic/nri-prometheus/pull/488 are valid, but the PR is showing that we have some issues in the pipeline.

This PR fixes both

### DIFF
--- a/charts/nri-prometheus/Chart.yaml
+++ b/charts/nri-prometheus/Chart.yaml
@@ -8,30 +8,24 @@ sources:
   - https://github.com/newrelic/nri-prometheus/tree/main/charts/nri-prometheus
 
 version: 2.1.18
-appVersion: 2.18.4
+appVersion: 2.21.4
 
 dependencies:
   - name: common-library
-    version: 1.1.1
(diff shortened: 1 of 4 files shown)


charts/nri-prometheus/Chart.yaml
```diff
@@ -7,7 +7,7 @@ sources:
   - https://github.com/newrelic/nri-prometheus
   - https://github.com/newrelic/nri-prometheus/tree/main/charts/nri-prometheus
 
-version: 2.1.17
+version: 2.1.18
 appVersion: 2.18.4
```
kang-makes, line 11:
```suggestion
appVersion: 2.21.4
```

kang-makes (commented):
Can we bump `nri-prometheus` image too?

charts/nri-prometheus/Chart.yaml
paologallinaharbur: sure, you are right!

kang-makes (approved):

PR #506, opened 2024-12-03
add fips compliant packages

rajrohanyadav:
## Summary
We need to support creating FIPS packages. For initial release we don't need FIPS images. Currently, `goreleaser` is being used to create and tag docker images for `nri-prometheus`. For FIPS, we need to create the binaries inside a container built using the centralized builder image: `ghcr.io/newrelic/coreint-automation:latest-go1.23.4-ubuntu16.04`.
We'll have to add `docker` and `buildx` dependencies to the image which are additional dependencies to maintain. It was tried with this [PR](https://github.com/newrelic/coreint-automation/pull/98).
I've tried to balance the mess and keep it only in this repo until we can fix it rather than adding unnecessary things to other repos.

***Note***: 
* Will close this [PR](https://github.com/newrelic/coreint-automation/pull/98) without merging.
* There are some code-duplication, but it makes sure we are not breaking anything for current customers in terms of image updates.
* Added a `TODO` to remove image creation from `goreleaser` and potentially use reusable workflow here.

## Testing
[Prerelease test](https://github.com/newrelic/nri-prometheus/actions/runs/12193255061)
image


.github/workflows/release-integration.yml
```diff
@@ -81,9 +81,17 @@ jobs:
...
+      - name: release fips
+        # FIPS packages need to be built in the fips container
+        # TODO: It should be refactored when we change the way docker images are being built
+        run: make ci/prerelease-fips
+        env:
+          GENERATE_PACKAGES: true
+          UPLOAD_PACKAGES: ${{ github.event.release.prerelease }}
+          GPG_MAIL: 'infrastructure-eng@newrelic.com'
+          # The upload script expects this to authenticate.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
paologallinaharbur, line 93: it seems a different approach than normal release. Why in why case is called "make release" and now "make ci/prerelease-fips"?
We have in one "release" and in the other "prerelease" 

Moreover, why for fibs we need as well GPG_MAIL and GITHUB_TOKEN?

paologallinaharbur: why is it not a step in "make release"?

.goreleaser-fips.yml
```diff
@@ -0,0 +1,33 @@
+---
+version: 2
+project_name: nri-prometheus
+builds:
+  - id: nri-prometheus-nix-fips
+    main: ./cmd/nri-prometheus/
+    binary: nri-prometheus
+    ldflags:
+      - -s -w -X github.com/newrelic/nri-prometheus/internal/integration.Version={{.Version}} #-X main.gitCommit={{.Commit}} -X main.buildDate={{.Date}}
```
paologallinaharbur, line 9: why that piece is commented out?

build/Dockerfile
```diff
@@ -1,18 +0,0 @@
-FROM golang:1.23.4-bookworm
```
paologallinaharbur, line 1: if we were not doing docker in docker how this was used?

build/ci.mk
```diff
@@ -15,21 +16,21 @@ ci/debug-container: ci/deps
...
-			$(BUILDER_TAG) make validate
+			$(BUILDER_IMAGE) make validate
 
 .PHONY : ci/test
 ci/test: ci/deps
 	@docker run --rm -t \
 			-v $(CURDIR):/go/src/github.com/newrelic/nri-$(INTEGRATION) \
 			-w /go/src/github.com/newrelic/nri-$(INTEGRATION) \
-			$(BUILDER_TAG) make test
+			$(BUILDER_IMAGE) make test
```
paologallinaharbur, line 33: all these ci/test, ci/validate, ci/snyck, ci/deps, ci/debug-container seems not used anyware. Can we remove them?

build/ci.mk
```diff
@@ -48,8 +49,29 @@ ifdef TAG
 			-w /go/src/github.com/newrelic/nri-$(INTEGRATION) \
 			-e INTEGRATION=$(INTEGRATION) \
 			-e TAG \
-			$(BUILDER_TAG) make release/build
+			$(BUILDER_IMAGE) make release/build
```
paologallinaharbur, line 52: also ci/build is used only in tests, I think we should test in the same way we release, I would remove this target to avoid confusion

paologallinaharbur: the only reason to keep it would be if in the close future we do docker in docker or if we change the way we build the image. But I do not think we are going to do it in the close future

build/ci.mk
```diff
@@ -48,8 +49,29 @@ ifdef TAG
...
+			-w /go/src/github.com/newrelic/nri-$(INTEGRATION) \
+			-e INTEGRATION \
+			-e PRERELEASE=true \
+			-e GITHUB_TOKEN \
+			-e REPO_FULL_NAME \
+			-e TAG \
+			-e TAG_SUFFIX \
+			-e GENERATE_PACKAGES \
+			-e UPLOAD_PACKAGES \
+			$(BUILDER_IMAGE) make release-fips
```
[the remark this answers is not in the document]
paologallinaharbur: that does both release and pre-release leveraging GENERATE_PACKAGES to decide 🫤 

Can we change GENERATE_PACKAGES name?
Can we  improve the names?

build/release.mk
```diff
@@ -31,11 +31,23 @@ ifeq ($(GENERATE_PACKAGES), true)
...
+endif
+
+.PHONY : release/build-fips
+release/build-fips: release/deps release/clean
+ifeq ($(GENERATE_PACKAGES), true)
+	@echo "===> $(INTEGRATION) === [release/build] PRERELEASE/RELEASE compiling fips binaries, creating packages, archives"
+	# TAG_SUFFIX should be set as "-pre" during prereleases
+	@$(GORELEASER_BIN) release --config $(CURDIR)/.goreleaser-fips.yml --skip=validate --clean
+else
+	@echo "===> $(INTEGRATION) === [release/build] build compiling fips binaries"
```
paologallinaharbur, line 48: Still confusing 😕 
On branch says `[release/build] PRERELEASE/RELEASE compiling fips binaries, creating packages, archives"`
the other says   `[release/build] build compiling fips binaries"`

build/release.mk
```diff
@@ -31,11 +31,23 @@ ifeq ($(GENERATE_PACKAGES), true)
...
+	@$(GORELEASER_BIN) build --config $(CURDIR)/.goreleaser.yml --skip=validate --snapshot --clean
+endif
+
+.PHONY : release/build-fips
+release/build-fips: release/deps release/clean
+ifeq ($(GENERATE_PACKAGES), true)
+	@echo "===> $(INTEGRATION) === [release/build] PRERELEASE/RELEASE compiling fips binaries, creating packages, archives"
+	# TAG_SUFFIX should be set as "-pre" during prereleases
+	@$(GORELEASER_BIN) release --config $(CURDIR)/.goreleaser-fips.yml --skip=validate --clean
+else
```
paologallinaharbur, line 47: this if-else condition looks useless, as far as I see GENERATE_PACKAGES is always true, also in the existing code! I think we should remove the dead code

build/release.mk
```diff
@@ -53,13 +65,26 @@ ifeq ($(UPLOAD_PACKAGES), true)
...
 endif
 	@echo "===> $(INTEGRATION) === [release/publish] publishing manifests"
-	@$(GORELEASER_BIN) build --config $(CURDIR)/.goreleaser.yml --skip-validate --snapshot --rm-dist
+	@$(GORELEASER_BIN) build --config $(CURDIR)/.goreleaser.yml --skip=validate --snapshot --clean
 
+.PHONY : release/publish-fips
+release/publish-fips:
+ifeq ($(UPLOAD_PACKAGES), true)
+	@echo "===> $(INTEGRATION) === [release/publish] publishing packages"
+	# REPO_FULL_NAME here is only necessary for forks. It can be removed when this is merged into the original repo
```
paologallinaharbur, line 74: is this an old and forgotten comment?

build/release.mk
```diff
@@ -53,13 +65,26 @@ ifeq ($(UPLOAD_PACKAGES), true)
 	@bash $(CURDIR)/build/upload_artifacts_gh.sh $(REPO_FULL_NAME)
 endif
 	@echo "===> $(INTEGRATION) === [release/publish] publishing manifests"
-	@$(GORELEASER_BIN) build --config $(CURDIR)/.goreleaser.yml --skip-validate --snapshot --rm-dist
+	@$(GORELEASER_BIN) build --config $(CURDIR)/.goreleaser.yml --skip=validate --snapshot --clean
 
+.PHONY : release/publish-fips
+release/publish-fips:
+ifeq ($(UPLOAD_PACKAGES), true)
```
paologallinaharbur, line 72: UPLOAD_PACKAGES could be safely change to PRE-RELEASE making everything clearer?

build/release.mk
```diff
@@ -53,13 +65,26 @@ ifeq ($(UPLOAD_PACKAGES), true)
...
+	@echo "===> $(INTEGRATION) === [release/publish] publishing packages"
+	# REPO_FULL_NAME here is only necessary for forks. It can be removed when this is merged into the original repo
+	@bash $(CURDIR)/build/upload_artifacts_gh.sh $(REPO_FULL_NAME)
+endif
+	@echo "===> $(INTEGRATION) === [release/publish] publishing manifests"
+	@$(GORELEASER_BIN) build --config $(CURDIR)/.goreleaser-fips.yml --skip=validate --snapshot --clean
```
paologallinaharbur, line 78: what is this doing?

build/Dockerfile
rajrohanyadav: I don't think it was being used anywhere.

.github/workflows/release-integration.yml
rajrohanyadav: Yes, it is different. As stated in the PR description, we need to build FIPS packages in docker. To not impact anything unless we plan to remove the image building from `goreleaser`, FIPS is released as a separate step.

> Moreover, why for fibs we need as well GPG_MAIL and GITHUB_TOKEN?

I think it was copy-paste, it might not be needed. I'll remove these.

> why is it not a step in "make release"?

The current flow was cleaning `dist` after each release and upload. To minimize the changes needed, it was implemented this way.

.goreleaser-fips.yml
rajrohanyadav: It was copy-paste. I think it can be removed. Will do. Thanks.

build/ci.mk
rajrohanyadav: I don't have enough background to comment for sure that they are not used. I'd assume these are useful for local development. But if your suggestion is to remove them, I can.

build/ci.mk
rajrohanyadav: > the only reason to keep it would be if in the close future we do docker in docker or if we change the way we build the image. But I do not think we are going to do it in the close future

On the contrary my assumption was that we'll be changing the way we build images in near future. Moreover, with go 1.24, we might not need to build FIPS images in docker and could just build on the runner. But that is still 3 months away. IMO, we should make `nri-prometheus` similar to other OHI release flows. Right now it is very confusing and messy. But would be difficult to block FIPS work because of this.

.goreleaser-fips.yml
rajrohanyadav: done

build/release.mk
rajrohanyadav: Changed echo statement, so that it makes sense.

build/release.mk
rajrohanyadav: done

build/release.mk
rajrohanyadav: Seems like it, removed it.

build/release.mk
rajrohanyadav: It is used for push/pr workflow where, by default it is `false` and only binaries are built.

build/release.mk
rajrohanyadav: changed echo statement

build/ci.mk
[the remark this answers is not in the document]
rajrohanyadav: > make ci/prerelease-fips is calling make release-fips. It is confusing

It is the same format being used in all the other repos: [Example](https://github.com/newrelic/nri-elasticsearch/blob/a91b203664a1f98b364cc4af9962c57253402ab5/build/ci.mk#L56)

build/ci.mk
paologallinaharbur: I suggest to remove them, they create only confusion. you can doublecheck that no one calls those make targets

build/release.mk
paologallinaharbur: are we calling make release?

.goreleaser-fips.yml
```diff
@@ -0,0 +1,33 @@
...
+      - >-
+        {{- if eq .Arch "arm64" -}}
+        CC=aarch64-linux-gnu-gcc
+        {{- end }}
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    tags:
```
paologallinaharbur, line 22: is this tag used somehow?

build/ci.mk
paologallinaharbur: what I meant is -> why do we have `ci/build` if in other parts we use directly `make release/build`

build/release.mk
```diff
@@ -31,11 +31,23 @@ ifeq ($(GENERATE_PACKAGES), true)
...
+
+.PHONY : release/build-fips
+release/build-fips: release/deps release/clean
+ifeq ($(GENERATE_PACKAGES), true)
+	@echo "===> $(INTEGRATION) === [release/build] PRERELEASE/RELEASE compiling fips binaries, creating packages, archives"
+	# TAG_SUFFIX should be set as "-pre" during prereleases
+	@$(GORELEASER_BIN) release --config $(CURDIR)/.goreleaser-fips.yml --skip=validate --clean
+else
+	@echo "===> $(INTEGRATION) === [release/build-fips] build compiling fips binaries"
+	# release/build with PRERELEASE unset is actually called only from push/pr pipeline to check everything builds correctly
```
paologallinaharbur, line 49: I think this comment is wrong or in the wrong place.
The if/else is checking GENERATE_PACKAGES, and we are not calling it set to false

paologallinaharbur (dismissed):

.goreleaser-fips.yml
rajrohanyadav: Yes, it is a go build tag to tell it to include `fips.go` and build for fips.

paologallinaharbur (approved):

sairaj18 (approved):

PR #509, opened 2024-12-11
Nr 341878 use reusable workflow for docker image security checks

rahulreddy15:
Successfully tested with the reusable workflow reusable_nightly.yaml.
PR needs to be merged after this https://github.com/newrelic/coreint-automation/pull/97 is merged.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -31,3 +31,4 @@ jobs:
       go_version_file: 'go.mod'
       trivy_scan: false
       generate_packages: true
+  
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -31,4 +31,3 @@ jobs:
       go_version_file: 'go.mod'
       trivy_scan: false
       generate_packages: true
- 
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -23,8 +23,7 @@ jobs:
       docker_tag: nightly
       target_branches: "main"
       integration_name: "prometheus"
-      build_command: |
-        make release
+      build_command: make release
       setup_qemu: true
       setup_buildx: true
       setup_go: true
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -23,6 +23,7 @@ jobs:
       docker_image: newrelic/nri-prometheus
       docker_tag: nightly
       target_branches: "main"
+      integration_name: "prometheus"
       build_command: |
         make release
       setup_qemu: true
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -1,45 +1,34 @@
 name: Nightly build
-
 on:
   schedule:
     - cron: "0 3 * * *"
   push:
     branches:
       - main
+  pull_request:
```
sigilioso, line 8: Are we supposed to create a `nightly` image on pull-requets or is it a leftover?

rahulreddy15: My bad. That was left over from testing...

.github/workflows/nightly.yaml
```diff
@@ -8,38 +7,27 @@ on:
...
+    secrets:
+      docker_username: ${{ secrets.FSI_DOCKERHUB_USERNAME }}
+      docker_password: ${{ secrets.FSI_DOCKERHUB_TOKEN }}
+    with:
+      docker_image: newrelic/nri-prometheus
+      docker_tag: nightly
+      target_branches: "main"
+      integration_name: "prometheus"
+      build_command: |
+        make release
```
rajrohanyadav, line 27: Why do we need it to be multiline when we just have a single line of input?

.github/workflows/nightly.yaml
```diff
@@ -8,38 +7,27 @@ on:
...
+      setup_qemu: true
+      setup_buildx: true
+      setup_go: true
```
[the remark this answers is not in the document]
rahulreddy15: Some don't need buildx and build  with a command. Buildx does more than building. Same with others.
There are exceptions so I had to include them.

.github/workflows/nightly.yaml
rahulreddy15: That was left over from copying configs from repos where multiline command were used.

rahulreddy15:
> I'm wondering why this change is required? Will we use this workflow in some other repos?

Yeah. All the integrations that release docker images will use it.

.github/workflows/nightly.yaml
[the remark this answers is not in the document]
rajrohanyadav: > Some don't need buildx and build with a command

AFAIK, `buildx` is needed when we need to build multi-arch images. And we mostly do that. I'm not sure when we wouldn't need it. But not a major thing.

.github/workflows/nightly.yaml
rajrohanyadav: Can we update it?

rajrohanyadav (approved):
LGTM

PR #556, opened 2026-03-26
fix: Pinned trivy action to 0.35.0

shashank-reddy-nr:
## Changes:
- Pinned trivy action version to 0.35.0.


vagrawal-newrelic (approved):
LGTM

